PR #3, opened 2020-02-19
Change package name capitalization

kaishin:
I know this is ballsy, but I would like to propose changing the package name to match the capitalization used in the official packages as well as other platform-agnostic 3rd party libraries. I understand if you'd rather stick to the convention used within iOS circles, in which case feel free to close this PR.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,7 +2,7 @@
 import PackageDescription
 
 let package = Package(
-  name: "CasePaths",
+  name: "case-paths",
   products: [
     .library(
       name: "CasePaths",


stephencelis:
Ha! I don't think we've ever noticed that 😅 Is there documentation around this guideline? And are there repercussions for making this change? Do you know how it affects downstream packages that depend on this one?

Package.swift
```diff
@@ -2,7 +2,7 @@
 import PackageDescription
 
 let package = Package(
-  name: "CasePaths",
+  name: "case-paths",
```
stephencelis: Should this be `swift-case-paths` per https://github.com/apple/swift-standard-library-preview/blob/833ea174a771bdc98edfa3529ae05439564e3902/Package.swift#L43?

kaishin: I wondered that too, but I decided to leave that decision up to you.

kaishin:
> Ha! I don't think we've ever noticed that 😅

Sometimes I feel like I am the only one who gets bothered by how divided the Swift community is when it comes to this one 😆

> Is there documentation around this guideline?

There isn't, and it's not a guideline. But dash-cased has been used more and more recently in official packages.

I have started [a discussion on the forum](https://forums.swift.org/t/swift-package-naming-conventions/) to see how the community feels about this.

> And are there repercussions for making this change?

I am not aware of any. The package is still pre-1.0 so I suppose it's relatively safe to assume that people expect things to change? For the record, I initially used this repo on a project, then replaced it with a fork with the name change and my project compiled without issues.

> Do you know how it affects downstream packages that depend on this one?

I highly doubt it. In the absence of a central repository, these kind of changes are largely insignificant. The product and target names are more critical, and those are unchanged in this PR.

stephencelis:
Given that even official packages (`SwiftPM` vs. `swift-nio`) are inconsistent, we'll keep an eye on the thread for some resolution 😄 Happy to make whatever change makes sense.

alexito4:
I thought SPM recently added some functionality that made things easier for the consumer if the package name was the same as the module. 🤔

stephencelis:
@kaishin We actually got around to changing this to `swift-case-paths` but forgot about this PR 😅 Sorry about that!

kaishin:
Thanks for taking care of this! I was just looking at it a couple of minutes ago and wondering if it's still relevant.

Issue #2, opened 2020-02-11
Add Cocoapods support


kirillsh:
Please add cocoapods support

stephencelis:
Hi @kirillsh! Thanks for taking the time to submit an issue. In order to better understand the need to support CocoaPods these days, is there anything that prevents you from using SPM to install this package into your project?

Dragna:
@stephencelis I think it maybe has to do with the fact that when a project is compatible with one dependency management system, you don't want to implement another one in order to minimize the friction of working with that project. Even if Xcode is doing a great work at hiding SPM in the development flow :)

In my team, we use Carthage for instance, we would like to use SPM, but the limitations are too much for us at the moment ( not possible to reference ressources :/ ) so we are sticking to it waiting for the update on that.

stephencelis:
@Dragna Could you still use SPM for any packages that don't need to package resources into a framework, etc.?

There are only a couple reasons I can think of to avoid SPM in an Xcode project these days:

1. The user cannot use Xcode 11 (I'm curious as to what might be preventing users from doing so, since it's been out for awhile and is a requirement for App Store submission)
2. The user already uses a different package manager and wants to use it for everything (but why would this prevent a user from using SPM or another package manager if a desired library doesn't support their manager of choice?)

We've supported CocoaPods and Carthage in all of our other libraries for many years, but we are using the release of CasePaths to reevaluate the need to support both of them, since SPM seems like a viable alternative these days. Maintaining support for CocoaPods and Carthage is a significant cost and the most brittle part of our ability to iterate and release new versions. Some things can be fixed with tooling, and the additional investment required in this kind of tooling feels significant.

If cases can be made for supporting CocoaPods and Carthage (that is, SPM is not even an option), and we have community members interested in managing this support, we're definitely open to it, but I don't think we yet understand why support may be needed when SPM seems to be an available alternative.

Dragna:
I could use SPM yes. and as a matter of fact that is how I will use it ^^. (it will help me make the continuous integration team of my company work on SPM integration 😛 ) 

I was just giving a possible explanation on why @kirillsh asked for cocoa pods support.

kirillsh:
thx @Dragna for the explanation.
The goal in our company is to have a unified solution for package management and we have it with pods. One more reason is SPM packages are not cached on our CI versus pod ones.

sergdort:
[...]

PR #4, opened 2020-02-26
CocoaPods Support

AdiAyyakad:
Love the pod and usage of it! Just wanted to create a quick little podspec so others can use it easily too :)

Please feel free to edit the authors and stuff I have it all pointed to my fork for now, but obviously would love it to just be part of this module as a first party thing.

### DIFF
--- a/CasePaths.podspec
+++ b/CasePaths.podspec
@@ -1,17 +1,20 @@
 Pod::Spec.new do |s|
-  s.name             = 'CasePaths'
-  s.version          = '0.1.0'
-  s.summary          = 'A short description of CasePaths.'
+  # Version
+  s.version          = "0.1.0"
+  s.swift_version    = "5.3"
 
-  s.description      = <<-DESC
-TODO: Add long description of the pod here.
-                       DESC
+  # Meta
+  s.name         = "CasePaths"
+  s.summary      = "Case paths bring the power and ergonomics of key paths to enums!"
+  s.license      = "MIT"
+  s.homepage     = 'https://github.com/AdiAyyakad/CasePaths'
+  s.authors      = { 'Aditya Ayyakad' => 'aditya.ayyakad@gmail.com' }
+  s.description  = <<-DESC
+  Case paths bring the power and ergonomics of key paths to enums!
+                   DESC
 
-  s.homepage         = 'https://github.com/AdiAyyakad/CasePaths'
-  s.author           = { 'Aditya Ayyakad' => 'aditya.ayyakad@gmail.com' }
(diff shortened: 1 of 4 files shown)


stephencelis:
Hey @AdiAyyakad! Thanks for taking the time to PR!

Could you look over the discussion at #2 to let us know if we're missing anything? It seems like CocoaPods support isn't necessary for this kind of library these days as SPM is an alternative.

AdiAyyakad:
Ahh I just read through that. I think it's really just as you noted option 2 where we have one dependency manager and don't want to use two. It's not a dealbreaker and honestly I've never used SPM in the past, so perhaps worth trying. We're just not going away from Objective-C in some of our codebase + dependencies anytime soon so doesn't seem worthwhile to invest in multiple dependency managers; especially when adding cocoapod support to another library is super simple as you can see. Not saying it's necessarily something pointfree needs to invest in - luckily that's the beauty of opensource :)

stephencelis:
Since SPM is built into Xcode we think it's a bit safer to get behind. And I definitely encourage you to give it a shot! It's pretty nice these days 😄

I'll keep this PR available for a bit in case we get more motivation discussed in #2, though!

AdiAyyakad:
Ahh so one thing I didn't note was that we're doing some version of composable architecture, definitely with less structure than your implementation (which has been so cool watching unfold!). However, we've already invested kind of heavily in doing it all via Cocoapods using pods and subspecs to build and distribute the modules internally. Without totally retooling, we actually need to use this spec ☝️ to even bring in CasePaths to our modules. Again, definitely the grave we've dug for ourselves but it is a big enough reason and again reasonable for anyone who doesn't have 100% swift dependencies like us.

ixrevo:
The problem with SPM arises when you have multiple transitive dependencies because SPM integrates its modules as static libraries. For example, I can't create an SPM module with ComposableArchitectre adapted to RxSwift, because
App -> RxDataSources
RxDataSources -> RxSwift
App -> RxComposableArchitecture
RxComposableArchitecture -> RxSwift
and I got two copies of RxSwift in the app. I was unsuccessful in finding a workaround for this problem. The support of Cocoapods would help.

AdiAyyakad:
@ixrevo I do have a working podspec up at `http://github.com/AdiAyyakad/swift-case-paths` if you want to just point to master directly 🤷‍♂ that's what we've been doing - would still be nice to get this in the main repo but it hasn't been a blocker for us

stephencelis:
@ixrevo I believe this is a bug with Xcode 11.4 :( 11.3 didn't have this problem. I've filed feedback with Apple but feel free to dupe!

Issue #17, opened 2020-12-08
Potential performance bottleneck when using reflection.


RabugenTom:
I'm reporting an issue I had with TCA. My app was lagging a lot and instruments showed that a bunch of the time was spent on the main thread computing debug description strings. This is caused by `Mirror(reflecting:)` which is used when trying to reassign local cases ([EnumReflection.swift#L61](https://github.com/pointfreeco/swift-case-paths/blob/469ef5103670d65f554f494ce9b6a46e0ec286db/Sources/CasePaths/EnumReflection.swift#L61)). 

I have huge data structs, with many containers and `NSAttributedString`, and generating the debug description doesn't come for free. The app is quite complex, but in the current state, ~80% of its work on the main thread is spent extracting `CasePath`'s. I temporarily improved the situation by carped-bombing my data structs with a some protocol extending `CustomReflectable` and providing a very small default implementation, but this is far from optimal.

I think that this kind of call is not expected to happen on the main thread, furthermore at a high frequency rate like it happens in a reducer.

Maybe the result can be cached, or another approach to reflection can be used, with some kind of (optional) protocol or prescription over `Action` enums?

mbrandonw:
Hey @RabugenTom,

The performance of case path reflection is a known issue, and it's definitely a bummer. We're not sure there is anything we can do to speed it up. Ideally Swift would provide first class support for this (like they do for key paths), which would make it much much faster.

We have found that reflection case paths work just fine as long as actions aren't being sent at a rapid pace. We're working on a large application right now and the reflection hasn't been an issue yet.

However, if you do have some actions that are being sent in great volume, then you may need to write your case paths by hand rather than relying on the reflection. There's a discussion in this issue on TCA discussing this very topic:

https://github.com/pointfreeco/swift-composable-architecture/pull/327

Let us know if that helps at all.

Also, I do not believe your performance problems are coming from usage of `.debug()` (or at least it shouldn't be!). By default `.debug()` uses a synchronous background queue to do printing, and so hopefully it shouldn't slow down your application: 

https://github.com/pointfreeco/swift-composable-architecture/blob/c1f88bd8608e8a26fd61ab46c9f11a63b0f4023d/Sources/ComposableArchitecture/Debugging/ReducerDebugging.swift#L158-L161

[...]

PR #18, opened 2020-12-10
Improve value extraction performance using `hashValue` when available

RabugenTom:
This pull request tries to improve the issue  #17.
While extracting the value from a case initializer using reflection, describing `any` as a String without precaution can cause performance issues when the value's description is very long, like it happens for large arrays or dictionaries for example.
In this cases, using a more compact marker is preferable. A common candidate is the hashValue of Hashable values. 
Thus, when `any` is Hashable, the library now uses its `hashValue` instead of the value itself as a path component. In other cases, it uses `any` description.

_Please note that if `hashValue` is now deprecated as Hashable requirement, it is very unlikely it will be ever removed. In any case, it can be reimplemented as_
```
extension Hashable {
    var hashValueCompat: Int {
        var hasher = Hasher()
        hasher.combine(self)
        return hasher.finalize()
    }
}
```
_This is the same implementation used in Swift's [Hashable.swift](https://github.com/apple/swift/blob/main/stdlib/public/core/Hashable.swift)._

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -67,7 +67,10 @@ public func extract<Root, Value>(case embed: (Value) -> Root, from root: Root) -
       any = anyChild
     }
     if MemoryLayout<Value>.size == 0, !isUninhabitedEnum(Value.self) {
-      return (["\(root)"], unsafeBitCast((), to: Value.self))
+      // If `any` is hashable, using its `hashValue` as path component is 
+      // likely faster than describing `any` itself.
+      let path = [String(describing: (any as? AnyHashable)?.hashValue ?? any)]
+      return (path, unsafeBitCast((), to: Value.self))
     }
     return nil
   }
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -67,8 +67,8 @@ public func extract<Root, Value>(case embed: (Value) -> Root, from root: Root) -
       any = anyChild
     }
     if MemoryLayout<Value>.size == 0, !isUninhabitedEnum(Value.self) {
-      // If `any` is hashable, using its `hashValue` as path component is 
-      // likely faster than describing `any` itself.
+      // If `root` is hashable, using its `hashValue` as path component is 
+      // likely faster than describing `root` itself.
       let path = [String(describing: (root as? AnyHashable)?.hashValue ?? root as Any)]
       return (path, unsafeBitCast((), to: Value.self))
     }
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -69,7 +69,7 @@ public func extract<Root, Value>(case embed: (Value) -> Root, from root: Root) -
     if MemoryLayout<Value>.size == 0, !isUninhabitedEnum(Value.self) {
       // If `any` is hashable, using its `hashValue` as path component is 
       // likely faster than describing `any` itself.
-      let path = [String(describing: (any as? AnyHashable)?.hashValue ?? any)]
+      let path = [String(describing: (root as? AnyHashable)?.hashValue ?? root as Any)]
       return (path, unsafeBitCast((), to: Value.self))
     }
     return nil
--- a/Tests/CasePathsTests/CasePathsTests.swift
+++ b/Tests/CasePathsTests/CasePathsTests.swift
@@ -364,6 +364,22 @@ final class CasePathsTests: XCTestCase {
       (/EnumWithLabeledCase.labeled(label:otherLabel:)).extract(
         from: .labeled(label: 2, otherLabel: 2)))
   }
+  
+  func testSiblingCases() {
+    enum Parent {
+      case foo(Foo)
+      case bar(Bar)
+    }
+    enum Foo {
+      case value(Int)
+    }
+    enum Bar {
+      case unrelated
+    }
+    XCTAssertNil(
+      extract(case: { Parent.bar(.unrelated) }, from: Parent.foo(.value(42)) )
+    )
+  }
 
   //  func testStructs() {
   //    struct Point { var x: Double, y: Double }


Sources/CasePaths/EnumReflection.swift
```diff
@@ -67,7 +67,10 @@ public func extract<Root, Value>(case embed: (Value) -> Root, from root: Root) -
       any = anyChild
     }
     if MemoryLayout<Value>.size == 0, !isUninhabitedEnum(Value.self) {
-      return (["\(root)"], unsafeBitCast((), to: Value.self))
```
stephencelis, line 70: This code path should only hit for enum cases with no associated data, e.g.:

```swift
enum Foo {
  case bar
}
```

It sounds like the stringified enum cases were more complex and had more data, right?

Is there a chance you can recreate a poorly-performing extraction? We could get a benchmark in place using [swift-benchmark](https://github.com/google/swift-benchmark) to get a better handle on how to improve performance.

stephencelis: Oh, I see, this code path also hits for an embedded class/struct in the deepest code path. Lemme get a benchmark in place to see if we can get a check for this.

stephencelis: Actually ha, this is still for things with zero memory, which wouldn't be most structs. We'd still really appreciate an example data type we can benchmark 😄

RabugenTom: Hello Stephen,

It hits when you're trying to match some case without associated value with another case with associated value from a sibling enum. For example, the following test passes, but hits L70 with `root` being 42:
```
func testSiblingCases() {
  enum Parent {
    case foo(Foo)
    case bar(Bar)
  }
  enum Foo {
    case value(Int)
  }
  enum Bar {
    case unrelated
  }
  
  XCTAssertNil(
    extract(case: { Parent.foo(.value(42)) }, from: Parent.bar(.unrelated))
  )
}
```

RabugenTom: I've just roughly tested the Hashable/nonHashable cases using a 100k chars string in `Foo.value` and the `Hashable` route is ~1000x faster. Proper benchmarks should show interesting results.

Please note that in order to enter the `Hashable` branch, `Parent` has to be `Hashable`. Automatic conformance induces that all child enums and their associated values should be `Hashable` too.

RabugenTom:
Hey @stephencelis,

Out of curiosity, and to practice what I've learnt recently, I  ran some proper benchmarks on my branch. I've tested failed extraction of a naked case from one with an `Int`, a 1000 chars `String`, a 1,000,000 chars `String`, and some custom nested struct of moderate complexity. 

Here are the results:
```
name                                        time            std        iterations
---------------------------------------------------------------------------------
Failed Extraction from Int.Any                 37589.000 ns ±  33.36 %      35716
Failed Extraction from Int.Hashable            20584.000 ns ±  35.07 %      65469

Failed Extraction from String.Any              72258.000 ns ±  17.42 %      17918
Failed Extraction from String.Hashable         21484.000 ns ±  69.94 %      65150

Failed Extraction from Long String.Any      37251677.000 ns ±   4.26 %         37
Failed Extraction from Long String.Hashable   413369.000 ns ±  10.91 %       3154

Failed Extraction from Container.Any        10246202.000 ns ±   6.91 %        131
Failed Extraction from Container.Hashable     261409.000 ns ±  11.50 %       5128
```

Fun fact: the `Int` case is faster with the `Hashable` route because it simply has to describe an `Int` (the `hashValue`), whereas the non-`Hashable` route as to describe `Parent2.Foo.int(42)`.

Please note that for moderate payloads, it can take around 10ms to fail to extract the very light `Parent.Bar.unrelated` case from the moderate struct, in the non-`Hashable` setup. In a TCA app where it happens on the main thread, that's almost 1 frame at 60fps each time the app fails to resolve an action. If the app is quite complex and if action traffic is dense (like it's the case when transitioning from screen to screen for example), this can cause dropped frames at the very best, up to an unresponsive app. This is what I've experienced in #17.

My PR assumes that `hashValues` collisions are rare and comparing them is enough to assert equality. I feel this should be the case for automatic conformance, but I may be wrong. It also should be the case if the user implements manually `Hashable` using the modern `hash(into:)` prescription. It can be problematic with inadequate `hashValue` implementations though.

Here is the benchmarking code I've used:
```
import Benchmark
import CasePaths
import Foundation

enum Parent1 {
  case foo(Foo)
  case bar(Bar)
}

enum Parent2: Hashable {
  case foo(Foo)
  case bar(Bar)
}

enum Foo: Hashable {
  case int(Int)
  case string(String)
  case container(Container)
}

enum Bar: Hashable {
  case unrelated
}

let id = UUID()
struct Container: Hashable {
  struct Item: Hashable {
    struct SubItem: Hashable {
      struct SubItem2: Hashable {
        var items: [UUID] = Array(repeating: id, count: 10)
      }
      var items: [SubItem2] = Array(repeating: SubItem2(), count: 10)
    }
    var items: [SubItem] = Array(repeating: SubItem(), count: 10)
  }
  var items: [Item] = Array(repeating: Item(), count: 10)
}

let intValueCasePathExtraction = BenchmarkSuite(name: "Failed Extraction from Int") { suite in
  suite.benchmark("Any") {
    extract(case: { Parent1.bar(.unrelated) }, from: Parent1.foo(.int(42)))
  }

  suite.benchmark("Hashable") {
    extract(case: { Parent2.bar(.unrelated) }, from: Parent2.foo(.int(42)))
  }
}

let string = String(repeating: "A", count: 1_000)
let stringValueCasePathExtraction = BenchmarkSuite(name: "Failed Extraction from String") { suite in
  suite.benchmark("Any") {
    extract(case: { Parent1.bar(.unrelated) }, from: Parent1.foo(.string(string)))
  }

  suite.benchmark("Hashable") {
    extract(case: { Parent2.bar(.unrelated) }, from: Parent2.foo(.string(string)))
  }
}

let longString = String(repeating: "A", count: 1_000_000)
let longStringValueCasePathExtraction = BenchmarkSuite(name: "Failed Extraction from Long String") { suite in
  suite.benchmark("Any") {
    extract(case: { Parent1.bar(.unrelated) }, from: Parent1.foo(.string(longString)))
  }

  suite.benchmark("Hashable") {
    extract(case: { Parent2.bar(.unrelated) }, from: Parent2.foo(.string(longString)))
  }
}

let container = Container()
let containerValueCasePathExtraction = BenchmarkSuite(name: "Failed Extraction from Container") { suite in
  suite.benchmark("Any") {
    extract(case: { Parent1.bar(.unrelated) }, from: Parent1.foo(.container(container)))
  }

  suite.benchmark("Hashable") {
    extract(case: { Parent2.bar(.unrelated) }, from: Parent2.foo(.container(container)))
  }
}

Benchmark.main([
  intValueCasePathExtraction,
  stringValueCasePathExtraction,
  longStringValueCasePathExtraction,
  containerValueCasePathExtraction,
])
```
Switching between `Parent1` and `Parent2` allows to test the non-`Hashable` and the `Hashable` routes. Both names have the same length so it should not cause any difference by itself.

mbrandonw:
Hey @RabugenTom, thanks for researching this and pushing us to improve the performance!

I'm still trying to understand everything you've done here. As far as I can tell this code path only seems to execute when dealing with a case path of the form `CasePath<Root, Void>`. For example, in our test suite, that code path is hit only on the following assertions:

```swift
    XCTAssertNotNil(
      extract(case: { Foo.bar }, from: .bar)
    )
    XCTAssertNil(
      extract(case: { Foo.bar }, from: .baz)
    )

    XCTAssertNotNil(
      extract(case: { Foo.baz }, from: .baz)
    )
    XCTAssertNil(
      extract(case: { Foo.baz }, from: .bar)
    )
```

Here `case` is a function `() -> Foo`.

This makes me believe that this change wouldn't drastically improve the performance of case paths used in most real world TCA applications because we aren't often dealing with `CasePath<Root, Void>` case paths. We usually pullback along case paths into sub-actions, such as `CasePath<AppAction, SettingsAction>`.

Am I understanding correctly? Were you able to measure any speed differences in your pulled back reducers with this change?

RabugenTom:
Hey @mbrandonw,

I've finally managed to reproduce the issue in a toy TCA app! When some naked child action produces, in the parent reducer, a synchronous `Effect` emitting some sibling action with an associated value, the emitted action goes through the problematic code. In a more schematic way, in the case:
```
ChildAction1.someAction
  ->rootReducer
    ->Effect(value: ChildAction2.someAction("Something")
```
In the toy app, `ChildAction1.someAction` is some `onAppear` action of some view, and `ChildAction2.someAction(...)` some action that sets the text of another view. 

By running the app, you can check that the action from the effect hits L70, and the associated value has to be described:
`textAction(TCAIssue.TextAction.setText("CURRENT VALUE, WILL BE UPPERCASED WHEN OTHERVIEW APPEARS"))`

I don't know if there are more simple examples, or if I'm doing things the wrong way. In my app, using the `hashValue` route  instead fixed a 1s main thread lag when pressing a button on some screen, so it can be very visible!

Maybe it can be fixed differently than using `Hashable` by detecting this specific configuration, but I'm not expert enough with TCA and CasePaths internals to answer.

Here is the toy app:
```
import ComposableArchitecture
import SwiftUI

struct RootState: Equatable {
  var textState = TextState(text: "Current value, will be uppercased when OtherView appears")
  var otherState = OtherState()
}

struct TextState: Equatable {
  var text: String
}

struct OtherState: Equatable {}

enum RootAction {
  case textAction(TextAction)
  case otherAction(OtherAction)
}

enum TextAction {
  case setText(String)
}

enum OtherAction {
  case onAppear
}

struct AppEnvironment {}

let otherReducer = Reducer<OtherState, OtherAction, AppEnvironment> {  _, _, _ in .none }

let textReducer = Reducer<TextState, TextAction, AppEnvironment> {
  state, action, _ in
  switch action {
  case let .setText(string):
    state.text = string
    return .none
  }
}

let rootReducer = Reducer<RootState, RootAction, AppEnvironment>.combine(
  otherReducer.pullback(state: \.otherState,
                        action: /RootAction.otherAction,
                        environment: { $0 }),
  textReducer.pullback(state: \.textState,
                       action: /RootAction.textAction,
                       environment: { $0 }),
  Reducer<RootState, RootAction, AppEnvironment> {
    state, action, _ in
    switch action {
    case .textAction:
      return .none
    case let .otherAction(action):
      switch action {
      case .onAppear:
        let updatedText = state.textState.text.uppercased()
        // Of course we can directly update the state here, but let
        // say we want to benefit from `setText` effects:
        return Effect(value: .textAction(.setText(updatedText)))
      }
    }
  }
)

struct RootView: View {
  let store: Store<RootState, RootAction>
  var body: some View {
    VStack {
      TextView(store: store.scope(state: \.textState, action: RootAction.textAction))
      OtherView(store: store.scope(state: \.otherState, action: RootAction.otherAction))
    }
    .frame(width: 300, height: 100)
  }
}

struct TextView: View {
  let store: Store<TextState, TextAction>
  var body: some View {
    WithViewStore(store) { viewStore in
      Text(viewStore.text)
    }
  }
}

struct OtherView: View {
  let store: Store<OtherState, OtherAction>
  var body: some View {
    WithViewStore(store) { viewStore in
      Color.red.frame(width: 44, height: 44)
        .onAppear { viewStore.send(.onAppear) }
    }
  }
}

struct RootView_Previews: PreviewProvider {
  static var previews: some View {
    RootView(store: .init(initialState: .init(), reducer: rootReducer, environment: .init()))
  }
}
```

stephencelis:
Hey @RabugenTom! We delved into this a bit more to try to better understand these reflection-based edge cases and think that the problem you're seeing only occurs when your app pulls actions back along single-case enums with no associated values. In your example:

```swift
enum OtherAction {
  case onAppear
}
```

If you change this enum to have a second case:

```swift
enum OtherAction {
  case onAppear
  case _placeholder
}
```

The inefficient code path in swift-case-paths shouldn't hit.

If this is the main way for this code path to hit (please let us know if you've found others!) I wonder if we can find a fix that doesn't require the end user to know they need to conform to `Hashable`, especially since adding `case _placeholder` is maybe less effort. We already have access to the enum type descriptor [here](https://github.com/pointfreeco/swift-case-paths/blob/469ef5103670d65f554f494ce9b6a46e0ec286db/Sources/CasePaths/EnumReflection.swift#L101-L138). Maybe we could make use of that to detect this situation and avoid the `String.init(describing:)` logic entirely.

We'll try to look into this soon because it's an edge case people can definitely hit, but we also think that the solution should not require the end use to conform their actions to `Hashable`. If you figure out another solution we'd love to see it!

RabugenTom:
Hey @stephencelis, thanks for your response.
In my app, I've transitioned to `Sourcery`-generated explicit `CasePath`'s, so reflection isn't used anymore. When the app will be finished, I'll switch back to reflection to see if other configurations are hitting this code with associated values.

I agree that `Hashable` can cause difficulties in some cases. It was a quick purely additive fix I've found that allowed me to continue developing my app.

I'll see if I find a workaround. I'm not an expert in low-level Swift internals, but I'll try to delve into it. In the meantime I'll close this PR as `Hashable` is off the table. I'll keep the source issue #17 open, but you can close it if you want.

PR #20, opened 2021-01-28
Fix memory leak due to mirror destructuring.

### LINKED ISSUE #19: Memory leak when using nested structs
To replicate build and profile the attached project.

Interestingly enough it doesn't leak memory if either or both of the properties are changed from `string` to `int`

[MemoryLeakTest.zip](https://github.com/pointfreeco/swift-case-paths/files/5876734/MemoryLeakTest.zip)

mbrandonw:
Looks like there's a Swift/`Mirror` bug that causes a leak if you destructure the tuple that is `mirror.children.first`. Luckily we can work around.

Fixes #19

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ test-linux:
 
 test-swift:
 	swift test \
-		--enable-pubgrub-resolver \
 		--enable-test-discovery \
 		--parallel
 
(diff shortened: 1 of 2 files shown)


stephencelis:
Opened https://bugs.swift.org/browse/SR-14121 to track this problem. Not sure an NB is necessary in the code.

stephencelis (approved):
Looks good! What a wild bug.

PR #29, opened 2021-05-26
Eliminate `/` ambiguity via disfavored overloads

stephencelis:
This changes `/` overloads to selectively use `@_disfavoredOverload`, which generally allows `/Foo.bar` to resolve to a case path without the ambiguity of key path-like "case path syntax" for extraction (like in `compactMap(/Result.success)`).

### DIFF
--- a/Sources/CasePaths/Operators.swift
+++ b/Sources/CasePaths/Operators.swift
@@ -32,11 +32,58 @@ public prefix func / <Root, Value>(
 ///   values. Its behavior is otherwise undefined.
 /// - Parameter root: A case with no an associated value.
 /// - Returns: A void case path.
-@_disfavoredOverload
 public prefix func / <Root>(
   root: Root
 ) -> CasePath<Root, Void> {
   .case(root)
+}
+
+/// Returns a case path for the given embed function.
+///
+/// - Note: This operator is only intended to be used with enum cases that have no associated
+///   values. Its behavior is otherwise undefined.
+/// - Parameter embed: An embed function.
+/// - Returns: A case path.
+public prefix func / <Root, A, B>(
+  embed: @escaping (A, B) -> Root
+) -> CasePath<Root, (A, B)> {
+  .case(embed)
+}
+
+/// Returns a case path for the given embed function.
+///
+/// - Note: This operator is only intended to be used with enum cases that have no associated
+///   values. Its behavior is otherwise undefined.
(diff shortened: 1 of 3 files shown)


Sources/CasePaths/Operators.swift
```diff
@@ -38,6 +38,54 @@ public prefix func / <Root>(
...
 }
 
+/// Returns a case path for the given embed function.
+///
+/// - Note: This operator is only intended to be used with enum cases that have no associated
+///   values. Its behavior is otherwise undefined.
+/// - Parameter embed: An embed function.
+/// - Returns: A case path.
+public prefix func / <Root, A, B>(
+  embed: @escaping (A, B) -> Root
```
stephencelis, line 48: Because of the `CasePath<Root, Void>` overload (so that `enum Foo { case bar }` can produce case path `/Foo.bar`), these overloads make case paths with multiple associated values more specific.

Without them:

```swift
enum Foo { case bar(Int, Int) }

/Foo.bar // assumes CasePath<(Int, Int) -> Foo, Void>, not CasePath<Foo, (Int, Int)>
```

Sources/CasePaths/Operators.swift
```diff
@@ -38,6 +38,54 @@ public prefix func / <Root>(
...
+  .case(embed)
+}
+
+/// Returns a case path for the given embed function.
+///
+/// - Note: This operator is only intended to be used with enum cases that have no associated
+///   values. Its behavior is otherwise undefined.
+/// - Parameter embed: An embed function.
+/// - Returns: A case path.
+public prefix func / <Root, A, B, C, D, E>(
```
stephencelis, line 83: Did overloads up to 5 associated values...maybe 3-4 is enough though?

Tests/CasePathsTests/CasePathsTests.swift
```diff
@@ -5,74 +5,73 @@ final class CasePathsTests: XCTestCase {
   func testEmbed() {
     enum Foo: Equatable { case bar(Int) }
 
-    XCTAssertEqual(.bar(42), (/Foo.bar).embed(42))
+    let fooBar = /Foo.bar
```
stephencelis, line 8: Almost all tests now pre-assign case paths via inference in order to ensure no ambiguity.

mbrandonw (approved):
nice!

PR #33, opened 2021-06-21
Cleanup + DocC

stephencelis:


.github/workflows/format.yml
```diff
@@ -11,11 +11,13 @@ jobs:
     runs-on: macOS-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Tap
+        run: brew tap pointfreeco/formulae
       - name: Install
-        run: brew install swift-format
+        run: brew install Formulae/swift-format@5.3
```
stephencelis, line 17: Need to fix formatting like we did on TCA.

Sources/CasePaths/EnumReflection.swift
```diff
@@ -1,5 +1,3 @@
-import func Foundation.memcmp
```
stephencelis, line 1: Not needed anymore after the last PR.

mbrandonw (approved):
awesome

PR #34, opened 2021-06-25
Conform CasePath to CustomStringConvertible

stephencelis:
Makes things a lil prettier when working with case paths in playgrounds, etc.


mbrandonw (approved):
Nice

PR #35, opened 2021-07-05
Improve reflection with Swift runtime and caching

stephencelis:
This code speeds up reflective case paths in a couple ways:

- We were previously constructing the "identity" of a case path's case by accumulating labels off a value's mirror, but we recently found some runtime metadata APIs that can compute the "tag" of an enum much more efficiently, so we can lean on that instead.
- Computing the enum tag can be done before we reflect on a value, so we can avoid that more intense work so long as we have previously successfully extracted a value from the case path. Upon the first successful extraction we can cache the case path's tag and use it to bail out early when future extractions don't match.

This PR also includes benchmarks that could be helpful when we make changes/improvements in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
-.swiftpm/
 /.build
 /Packages
 /*.xcodeproj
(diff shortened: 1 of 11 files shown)


mbrandonw (approved):
Sweet

PR #36, opened 2021-07-05
use destructiveProjectEnumData value witness function

mayoff:
The destructiveProjectEnumData function in the enum value witness table
transforms an enum value into its associated values in place, by zeroing
out any enum tag bits stored in the payload area of the value.

We can use this function after verifying that the tag of the enum
value is the expected tag. This means we still have to use the
Mirror-and-embed strategy until we can cache the expected tag.

Performance without this commit, on my iMac Pro:

```
name               time        std        iterations
----------------------------------------------------
Success.Manual       34.000 ns ± 232.27 %    1000000
Success.Reflection 3095.000 ns ±  22.64 %     470183
Failure.Manual       36.000 ns ± 167.71 %    1000000
Failure.Reflection   83.000 ns ± 152.31 %    1000000
```

Performance with this commit:

```
name               time       std        iterations
---------------------------------------------------
Success.Manual      35.000 ns ± 158.00 %    1000000
Success.Reflection 167.000 ns ±  70.60 %    1000000
Failure.Manual      37.000 ns ± 197.47 %    1000000
Failure.Reflection  82.000 ns ± 135.63 %    1000000
```

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -73,7 +73,7 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
 
     if let cachedTag = cachedTag {
       guard cachedTag == rootTag else { return nil }
-      return associatedValues(of: root)
+      return .some(associatedValues(of: root))
     }
 
     guard
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -146,7 +146,6 @@ private func associatedValues<Enum, Values>(of `enum`: Enum) -> Values {
   enumPtr.initialize(to: `enum`)
   let untypedPtr = UnsafeMutableRawPointer(enumPtr)
   vwt.destructiveProjectEnumData(untypedPtr, metadataPtr)
-  // Maybe this should be .bindMemory instead of .assumingMemoryBound?
   let valuesPtr = untypedPtr.assumingMemoryBound(to: Values.self)
   defer {
     valuesPtr.deinitialize(count: 1)
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -288,7 +288,7 @@ private struct EnumMetadata: Metadata {
 @_silgen_name("swift_allocBox")
 private func swift_allocBox(for metadata: UnsafeRawPointer) -> BoxPair
 
-@_silgen_name("swift_allocBox")
+@_silgen_name("swift_deallocBox")
 private func swift_deallocBox(_ heapObject: UnsafeMutableRawPointer)
 
 @_silgen_name("swift_projectBox")
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -85,18 +85,8 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
     guard
       let value = (Mirror(reflecting: root).children.first?.value)
         .flatMap({ $0 as? Value ?? Mirror(reflecting: $0).children.first?.value as? Value })
+        ?? workaroundForSR12044(Value.self)
     else {
-      #if compiler(<5.2)
-        // https://bugs.swift.org/browse/SR-12044
-        if
-          MemoryLayout<Value>.size == 0,
-          let valueMetadata = EnumMetadata(Value.self),
-          valueMetadata.kind == .enumeration,
-          valueMetadata.typeDescriptor.isInhabited
-        {
-          return unsafeBitCast((), to: Value.self)
-        }
-      #endif
       return nil
     }
     let embedTag = metadata.tag(of: embed(value))
@@ -302,18 +292,6 @@ private struct EnumTypeDescriptor {
     return FieldDescriptor(
       ptr: ptr.advanced(by: 4 * 4).loadRelativePointer())
   }
-
-  #if compiler(<5.2)
-  var numPayloadCases: Int32 {
-    return ptr.advanced(by: 5 * 4).load(as: Int32.self) & 0xFFFFFF
-  }
-
-  var numEmptyCases: Int32 {
-    return ptr.advanced(by: 6 * 4).load(as: Int32.self)
-  }
-
-  var isInhabited: Bool { numPayloadCases + numEmptyCases > 0 }
-  #endif
 }
 
 private struct FieldDescriptor {
@@ -344,3 +322,46 @@ extension FieldRecord {
     static var isIndirectCase: Self { .init(rawValue: 1) }
   }
 }
+
+#if compiler(<5.2)
+
+// https://bugs.swift.org/browse/SR-12044
+private func workaroundForSR12044<Value>(_ type: Value.Type) -> Value? {
+  // If Value is an inhabited type with a size of zero, Mirror doesn't notice it as the associated value of an enum case due to incorrect metadata. But, being inhabited with a size of zero, it has only one possible inhabitant, so I create the inhabitant here using unsafeBitCast.
+
+  // An uninhabited type like Never also has a size of zero, and I have to be careful not to create a value of an uninhabited type.
+
+  // It's possible for a tuple, struct, or class to be uninhabited by having an uninhabited stored property. But detecting such a type is difficult as I'd have to scrounge through even more metadata. So instead I'm just checking for the common case of an uninhabited enum. If you do something like `enum E { case c(Never, Never) }`... you have my sincere apology.
+
+  if
+    MemoryLayout<Value>.size == 0,
+    !isUninhabitedEnum(Value.self)
+  {
+    return unsafeBitCast((), to: Value.self)
+  }
+  return nil
+}
+
+private func isUninhabitedEnum(_ type: Any.Type) -> Bool {
+  // If it lacks enum metadata, it's definitely not an uninhabited enum.
+  guard let metadata = EnumMetadata(type) else { return false }
+  return !metadata.typeDescriptor.isInhabited
+}
+
+extension EnumTypeDescriptor {
+  var numPayloadCases: Int32 {
+    return ptr.advanced(by: 5 * 4).load(as: Int32.self) & 0xFFFFFF
+  }
+
+  var numEmptyCases: Int32 {
+    return ptr.advanced(by: 6 * 4).load(as: Int32.self)
+  }
+
+  var isInhabited: Bool { numPayloadCases + numEmptyCases > 0 }
+}
+
+#else
+
+private func workaroundForSR12044<Value>(_ type: Value.Type) -> Value? { nil }
+
+#endif
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -324,7 +324,6 @@ private struct FieldDescriptor {
   /// The size of a FieldRecord as stored in the executable.
   var recordSize: Int { Int(ptr.advanced(by: 2 * 4 + 2).load(as: UInt16.self)) }
 
-  var fieldCount: UInt32 { ptr.advanced(by: 2 * 4 + 2 * 2).load(as: UInt32.self) }
 
   func field(atIndex i: UInt32) -> FieldRecord {
     return FieldRecord(
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -100,7 +100,6 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
       #endif
       return nil
     }
-    if rootTag == cachedTag { return value }
     let embedTag = metadata.tag(of: embed(value))
     cachedTag = embedTag
     if rootTag == embedTag { return value }
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -1,3 +1,5 @@
+import CCasePaths
+
 extension CasePath {
   /// Returns a case path that extracts values associated with a given enum case initializer.
   ///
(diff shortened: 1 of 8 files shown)


Sources/CasePaths/EnumReflection.swift
```diff
@@ -131,19 +136,43 @@ private func isUninhabitedEnum(_ type: Any.Type) -> Bool {
...
+}
+
+private func associatedValues<Enum, Values>(of `enum`: Enum) -> Values {
+  let (metadataPtr, vwt) = enumBits(for: Enum.self)!
+  let enumPtr = UnsafeMutablePointer<Enum>.allocate(capacity: 1)
+  enumPtr.initialize(to: `enum`)
+  let untypedPtr = UnsafeMutableRawPointer(enumPtr)
+  vwt.destructiveProjectEnumData(untypedPtr, metadataPtr)
+  // Maybe this should be .bindMemory instead of .assumingMemoryBound?
+  let valuesPtr = untypedPtr.assumingMemoryBound(to: Values.self)
```
stephencelis, line 150: It think what you have with `assumingMemoryBound(to:)` is the right call. I do wonder if it's possible to do fewer raw-typed pointer dances by describing the value witness table slightly differently, but I'm not sure if that is safe to do.

```suggestion
  let valuesPtr = untypedPtr.assumingMemoryBound(to: Values.self)
```

stephencelis (approved):
Wow this is awesome! We were aware of that function but got stuck when we tried to use it to completely replace the `Mirror` reflection code. It didn't occur to us that we could at the very least leverage this functionality safely once we had a tag cached 😅

If you're curious, we [have a branch going](https://github.com/pointfreeco/swift-case-paths/compare/echo) that brought in [Echo](https://github.com/Azoy/Echo) to make it easier to explore all of the runtime's corners. We're not sure if there's a material difference in our extraction code, but we leveraged a couple other metadata functions during the extraction that I think are equivalent to your manual `.allocate`:

https://github.com/pointfreeco/swift-case-paths/compare/echo#diff-42ec48eb0f16775a34c14e246bc584b400aef21d43fed56a8b52aab59589852eR89-R95

In the end we were unable to figure out how to safely avoid the `Mirror` code, and were bringing in a lot more runtime than we would have liked: https://github.com/pointfreeco/swift-case-paths/blob/d1af52f98304706156351bb3052e556c871051a2/Sources/CasePaths/EnumReflection.swift#L112-L150

That said, if anyone figures out how to avoid the reflection entirely, that'd be a nice improvement for any cold extraction, but this is a really massive improvement and you've narrowed the performance gap entirely for the cached path :smile:

stephencelis:
@mayoff Cool if I merge this? Just wanted to make sure you weren't still working on anything :smile:

mayoff:
I didn't have anything else to add but I'll be happy to investigate the problems when I get your test cases.

stephencelis:
@mayoff Just pushed a bunch of cases, where 3 fail:

- `testZeroMemoryLayoutPayload`
- `testRecursivePayload`
- `testOptionalPayload`

There may be other edge cases I haven't thought to test.

Swift's blog post on "How Mirror Works" has some helpful info on how Mirror's use the runtime APIs to extract a value: https://swift.org/blog/how-mirror-works/

In particular, indirect cases need to be extracted differently, which is why the recursive payload is failing:

> In the indirect case, the real data has to be pulled out of the box:

I haven't yet dug into the other two failures.

There's some very old Echo code that recreates the `Mirror` APIs using the Swift runtime _in_ Swift: https://github.com/Azoy/Echo/blob/f3ba95c4f9eeb72a63c84dc6f944fe1852990df9/Sources/EchoMirror/ReflectionMirror.swift

This is part of what we pulled from in our experiments, if that helps!

Anyway, hopefully this isn't too much of a rabbit hole (from personal experience we think it might be). If you're interested in figuring out these details we would absolutely love to help you get this merged!

mayoff:
Regarding the failure in `testExtract`, the failing assertion is

```swift
    XCTAssertEqual(
      [1],
      [Result.success(1), .success(nil), .failure(MyError())]
        .compactMap(/{ .success(.some($0)) })
    )
```

This calls into the `extract` method whose documentation says

> Note: This function is only intended to be used with enum case initializers. Its behavior is otherwise undefined.

but `{ .success(.some($0)) }` is not an enum case initializer. It's not clear to me that this is a legitimate test case.

mayoff:
I'm seeing a failure in `testExtract` but no failure in `testZeroMemoryLayoutPayload`. I've tried both Xcode 12.5.1 and Xcode 13 beta 2 on an Intel Mac running macOS 11.4 (20F71).

stephencelis:
> I'm seeing a failure in `testExtract` but no failure in `testZeroMemoryLayoutPayload`. I've tried both Xcode 12.5.1 and Xcode 13 beta 2 on an Intel Mac running macOS 11.4 (20F71).

You're absolutely right. That is a stale assertion that got snuck back in. Just removed it!

mayoff:
I'm still working on fixing the indirect problem.

mayoff:
All test cases now pass.

mayoff:
I should clarify that all test cases pass for me, in Xcode 12.5.1 on macOS 11.4.

Sources/CasePaths/EnumReflection.swift
```diff
@@ -93,57 +110,240 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
...
+    guard valueWitnessTable.flags.contains(.isNonInline) else { return }
+    swift_deallocBox(buffer.load(as: UnsafeMutableRawPointer.self))
+  }
+}
+
+@_silgen_name("swift_allocBox")
+private func swift_allocBox(for metadata: UnsafeRawPointer) -> BoxPair
+
+@_silgen_name("swift_allocBox")
+private func swift_deallocBox(_ heapObject: UnsafeMutableRawPointer)
```
stephencelis, line 292: Should the silgen name match here? 
```suggestion
@_silgen_name("swift_deallocBox")
private func swift_deallocBox(_ heapObject: UnsafeMutableRawPointer)
```

Sources/CasePaths/EnumReflection.swift
```diff
@@ -93,57 +110,240 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
...
-  // https://github.com/apple/swift/blob/master/stdlib/public/core/ReflectionMirror.swift
-  let isEnum = metadataKind == 0x201
-  guard isEnum else { return false }
+  /// The size of a FieldRecord as stored in the executable.
+  var recordSize: Int { Int(ptr.advanced(by: 2 * 4 + 2).load(as: UInt16.self)) }
 
-  // Access enum type descriptor
-  let enumMetadata = metadataPtr.load(as: EnumMetadata.self)
-  let enumTypeDescriptor = enumMetadata.typeDescriptor.pointee
+  var fieldCount: UInt32 { ptr.advanced(by: 2 * 4 + 2 * 2).load(as: UInt32.self) }
```
stephencelis, line 327: Looks like this isn't used?

Sources/CasePaths/EnumReflection.swift
```diff
@@ -69,22 +71,37 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
+          let valueMetadata = EnumMetadata(Value.self),
+          valueMetadata.kind == .enumeration,
+          valueMetadata.typeDescriptor.isInhabited
+        {
           return unsafeBitCast((), to: Value.self)
         }
       #endif
       return nil
     }
     if rootTag == cachedTag { return value }
```
stephencelis, line 103: Not sure if it was possible to hit this code path before, but definitely looks impossible now :smile:
```suggestion
```

Sources/CasePaths/EnumReflection.swift
```diff
@@ -93,57 +110,240 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
...
+
+    let (heapObject: heapObject, buffer: boxBuffer) = swift_allocBox(for: ptr)
+    buffer.storeBytes(of: heapObject, as: UnsafeMutableRawPointer.self)
+    return boxBuffer
+  }
+
+  func deallocateBoxForExistential(in buffer: UnsafeMutableRawPointer) {
+    guard valueWitnessTable.flags.contains(.isNonInline) else { return }
+    swift_deallocBox(buffer.load(as: UnsafeMutableRawPointer.self))
+  }
```
stephencelis, line 285: These functions appear to be only called in the code path for indirect cases, which has already guarded against the same condition, so I think these guards will always early out. This means the bottom half of each function is not currently exercisable. Do you think that means there's an opportunity to simplify (get rid of `@_silgen_name` references that aren't used, these code paths, etc.)?

stephencelis (commented):
🥳 Awesome! Things are all passing for me, too!

I followed through the code and found a few small things/questions.

We'd love to know the process you took to get to here if you have a chance to write anything up! I'm sure others would appreciate it, too :smile:

Sources/CasePaths/EnumReflection.swift
mayoff: lol yes

Sources/CasePaths/EnumReflection.swift
mayoff: Indeed it is not used because I just assume the enum tag is a valid index into the FieldRecords. The C++ code also makes this assumption.

Sources/CasePaths/EnumReflection.swift
mayoff: Yeah, that can be removed.

Sources/CasePaths/EnumReflection.swift
```diff
@@ -69,22 +71,36 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
       let value = (Mirror(reflecting: root).children.first?.value)
         .flatMap({ $0 as? Value ?? Mirror(reflecting: $0).children.first?.value as? Value })
     else {
       #if compiler(<5.2)
         // https://bugs.swift.org/browse/SR-12044
-        if MemoryLayout<Value>.size == 0, !isUninhabitedEnum(Value.self) {
+        if
+          MemoryLayout<Value>.size == 0,
+          let valueMetadata = EnumMetadata(Value.self),
+          valueMetadata.kind == .enumeration,
```
stephencelis, line 95: This can also be removed, right? Does `EnumMetadata` make that assumption?
```suggestion
```

Sources/CasePaths/EnumReflection.swift
mayoff: I kind of wondered that myself, but I thought maybe with a large enough associated value it would get triggered. And indeed it does. I've added test cases that bring EnumReflection.swift to 100% code coverage.

Sources/CasePaths/EnumReflection.swift
mayoff: `EnumMetadata` checks that the metadata kind is either `.enumeration` or `.optional` (they use the same metadata format), but this workaround code (prior to my changes) checked specifically for the `.enumeration` kind.

I'm not entirely sure what checks are needed in this workaround and didn't want to download an older Xcode to play around with it.

mayoff:
If you'd like me to squash the commits, let me know.

stephencelis:
> If you'd like me to squash the commits, let me know.

Not necessary 😄 We'll squash merge!

mayoff:
A couple of other thoughts:

- It would be reasonable for the `extract` method to call `preconditionFailure` if given a non-enum `Root` type. Right now it returns an extractor that always returns nil.

- In theory we could detect that bogus case path `/{ .success(.some($0)) }` by using the metadata to check that the case's associated value's type is the same as the user-provided `Value` type. In that bogus case, they wouldn't match because the `Value` type is `Int` but the associated value's type is `Optional<Int>`.

  `Mirror` has to figure out the associated value's true type in order to put it in an `Any` existential, but it does so using a function (`swift_getTypeByMangledName`) that isn't exposed (AFAIK) as ABI. There are some variants that are exposed (`swift_getTypeByMangledNameInContext` and `swift_getTypeByMangledNameInEnvironment`), but I haven't investigated how to use them and I believe there are no example uses in the Swift repo.

mayoff:
Looks like I'll be downloading Xcode 11.7 to work around that compiler crash…

For the Ubuntu failure, I'll just remove the `__nonnull` specifier and force-unwrap it instead.

stephencelis:
> * It would be reasonable for the `extract` method to call `preconditionFailure` if given a non-enum `Root` type. Right now it returns an extractor that always returns nil.

I think this makes sense, but maybe we can save the change for another PR/discussion.

> * In theory we could detect that bogus case path `/{ .success(.some($0)) }` by using the metadata to check that the case's associated value's type is the same as the user-provided `Value` type. In that bogus case, they wouldn't match because the `Value` type is `Int` but the associated value's type is `Optional<Int>`.
>   `Mirror` has to figure out the associated value's true type in order to put it in an `Any` existential, but it does so using a function (`swift_getTypeByMangledName`) that isn't exposed (AFAIK) as ABI. There are some variants that are exposed (`swift_getTypeByMangledNameInContext` and `swift_getTypeByMangledNameInEnvironment`), but I haven't investigated how to use them and I believe there are no example uses in the Swift repo.

Hm, are you suggesting a way of eliminating the `Mirror` code and extracting the value directly using the Swift runtime? Or are you talking about supporting that one bogus case path I had a test case for? The bogus/nested case path test was written back when we supported nested reflection `/{ .my(.deeply(.nested(.case($0)))) }`. We decided to get rid of support for that in #32, since we should always prefer things like `/A.my .. /B.deeply .. /C.nested .. /D.case`.

mayoff:
I was only suggesting a way of detecting when a bogus embed function is passed to the reflection-based extract function. It wouldn't enable support for them. It would just fail (crash) instead of silently returning the wrong answer. But it would be a lot of work (research) to implement.

Sources/CasePaths/EnumReflection.swift
mayoff: Since I had to download Xcode 11.3 to investigate the CI failure, I was able to test this and discover that I broke the workaround. It's fixed as of dd0a6d8.

stephencelis:
✅ @mayoff Things are looking good!

We're going to spend some time exercising this against various code bases tomorrow, and if nothing comes up it'll get an enthusiastic merge!

mayoff:
It definitely needs to be exercised on an arm64e platform with pointer authentication.

stephencelis:
I gave things a whirl today on a bunch of repos and a few devices today and things seemed to work without a problem! @mbrandonw has an arm64e he's gonna try soon, and if his experience is the same we should be good to merge. Thanks again!

iampatbrown:
This is great. Was testing it out yesterday and it was solid. Hadn't thought about about bypassing the reflection after verifying the embed tag. Awesome work @mayoff!

mbrandonw:
> It definitely needs to be exercised on an arm64e platform with pointer authentication.

Just ran the test suite on my iPhone XS Max as well as all the case studies and sample apps in the TCA repo and everything seems to be good! We will be merging soon 🎉

PR #37, opened 2021-07-08
eliminate the use of Mirror

mayoff:
This branch is not ready for merging. I'm creating this pull request only to make it available for discussion.

I found example uses of `swift_getTypeByMangledNameInContext` in [the Reflect package](https://github.com/wickwirew/Runtime) and have made some progress toward using it to replace the use of `Mirror`.

In its current state, this branch has several failing tests, all related to associated value labels. I've outlined what's wrong and what needs to be done in the commit message and code comments.

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -69,21 +69,20 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
 /// - Parameter embed: An enum case initializer.
 /// - Returns: A function that can attempt to extract associated values from an enum.
 public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -> (Value?) {
-  var cachedTag: UInt32?
-  return { root in
-    guard let metadata = EnumMetadata(Root.self) else { return nil }
-    let rootTag = metadata.tag(of: root)
+  guard EnumMetadata(Root.self) != nil else {
+    #if DEBUG
+    print("\(#function) (#file:#line) can never extract values from \(Root.self) because \(Root.self) isn't an enum!")
+    #endif
+    return { _ in nil }
+  }
 
-    if let cachedTag = cachedTag {
-      guard
-        cachedTag == rootTag,
-        let fieldDescriptor = metadata.typeDescriptor.fieldDescriptor
-        else { return nil }
-      return .some(
-        fieldDescriptor.field(atIndex: rootTag).isIndirectCase
-          ? metadata.indirectAssociatedValue(of: root, as: Value.self)
-          : metadata.directAssociatedValue(of: root, as: Value.self))
+  var cachedExtractor: Extractor<Root, Value>? = nil
+  return { root in
+    if let extractor = cachedExtractor {
+      return extractor.extract(from: root)
     }
+
+    let metadata = EnumMetadata(assumingEnum: Root.self)
 
     guard
       let value = (Mirror(reflecting: root).children.first?.value)
@@ -92,10 +91,16 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
     else {
       return nil
     }
+
     let embedTag = metadata.tag(of: embed(value))
-    cachedTag = embedTag
-    if rootTag == embedTag { return value }
-    return nil
+    let extractor = Extractor<Root, Value>(tag: embedTag)
+    cachedExtractor = extractor
+    // We could do this to avoid extracting the value again:
+    //
+    // return metadata.tag(of: root) == embedTag ? value : nil
+    //
+    // But for now I'm extracting a second time (but only the first time I find the tag I'm looking for!) to ensure that every test case that returns a value exercises the extractor.
+    return extractor.extract(from: root)
   }
 }
 
@@ -106,7 +111,7 @@ private let pointerSize = MemoryLayout<UnsafeRawPointer>.size
 
 private typealias OpaqueExistentialContainer = (
   UnsafeMutableRawPointer?, UnsafeMutableRawPointer?, UnsafeMutableRawPointer?,
-  metadata: UnsafeMutableRawPointer?
+  metadata: UnsafeRawPointer?
 )
 
 private typealias BoxPair = (heapObject: UnsafeMutableRawPointer, buffer: UnsafeMutableRawPointer)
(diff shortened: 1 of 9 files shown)


Sources/CasePaths/EnumReflection.swift
```diff
@@ -98,6 +119,36 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
...
+
+  // Consider these: `enum F { case d(a: Int, b: String) }`
+  //
+  // Certainly we should allow `CasePath<F, (a: Int, b: String)>` to match this.
+  //
+  // We would also like `CasePath<F, (Int, String)> to match this.
+  //
+  // So if `payloadType` is a tuple, and `desiredType` is a tuple with no labels, and the two types have identical elements types, I treat them as equivalent.
+
+  // TODO
```
stephencelis, line 147: We explored handling this stuff in another branch, and particularly a couple more cases here: https://github.com/pointfreeco/swift-case-paths/blob/20d0a0668731d1f59dec45ea115e980e32feac47/Sources/CasePaths/EnumReflection.swift#L132-L141

Protocol conformances might be tough, but maybe there's a simpler way we didn't know about.

mayoff: I don't understand the “handle protocol conformances” case.

Consider

```
protocol P { }
extension Int: P { }
enum E {
    case p(P)
    case i(Int)
}
```

Then the only legal `embed` functions we can pass to `extract` are these:

- `E.p`, which takes a `P` (existential) argument, and the returned extractor's type is `(E) -> P?`
- `E.i`, which takes an `Int` argument, and the returned extractor's type is `(E) -> Int?`

In neither case does the extractor need to convert a `P` existential to an `Int` or vice versa. Am I overlooking something?

stephencelis: In the above branch there is an assertion that fails if that `case let (fieldType as ExistentialMetadata...)` check is omitted:

```swift
    struct MyError: Equatable, Error {}

    XCTAssertEqual(
      .some(MyError()),
      (/Result<String, Error>.failure)
        .extract(from: .failure(MyError()))
    )
```

This is because the case path's `Value.self` is the `Error` existential, but the type we get from `swift_getTypeByMangledNameInContext` is `MyError`. I'm not sure if there's a simple means of making this `MyError is Error` test dynamically, which is why we have the more complicated metadata tests above.

stephencelis: (You could argue that this kind of case path is rare and that Swift's type system limitations means we shouldn't bother supporting it, but I wanted to note that it _does_ currently work with reflection and the metadata code from your last PR.)

mayoff: Okay, I see what's going on.

Swift infers a case path type of `CasePath<Result<String, Error>, MyError>`. It has to, because `XCTAssertEqual` requires `Equatable` conformance, and `MyError` is `Equatable` but `Error` is not.

Without constraints, the type of `Result<String, Error>.failure` is `(Error) -> Result<String, Error>`. But here, Swift needs the type to be `(MyError) -> Result<String, Error>`.

Because `Error` is in a contravariant position, `(Error) -> Result<String, Error>` is a subtype of `(MyError) -> Result<String, Error>`. Swift implicitly converts the function to the supertype (by wrapping it in a “reabstraction thunk helper”) to make the expression type-check. How gracious.

So we have an expression that looks fine syntactically (you're apparently passing an enum case initializer, as required), but  where the `Value` type is not the associated value's type.

Yes, I think we should support this, but perhaps just by falling back to `Mirror` and a `#if DEBUG` warning.

mayoff: It turns out my last PR also suffers from this bug. I'm working on it.

stephencelis: @mayoff Interesting, we probably didn't have good enough coverage on that. Did you cook up a failing test on `main`?

I updated this test and it still passes, though the case paths are pretty explicit in the types they return: https://github.com/pointfreeco/swift-case-paths/commit/1e224002a8a10ff8299dc0af72f82502150f2cc0

It's the end of the week, so energy is running a lil low, but it seems like even if your PR has this issue, I'm having a hard time thinking of a real world instance of it.

mayoff: Here's a failing test case:

https://github.com/mayoff/swift-case-paths/blob/41784e12a432daae9b4c23557b3796fc3af5db9a/Tests/CasePathsTests/CasePathsTests.swift#L361-L381

You'll also need the declarations related to `TestProtocol` from the top of the file.

mayoff:
The new version ([7cab867](https://github.com/pointfreeco/swift-case-paths/pull/37/commits/7cab86738ddb2be55561a91241a54085a534981b)) fixes the existential bug and passes all the tests (including some new ones).

Code coverage of `EnumReflection.swift` indicates the tests cover everything except an “impossible” case where some metadata is missing, and a mangled type name encoding that I don't know how to trigger.

It also runs faster.

Main branch ([90e4e67e](https://github.com/pointfreeco/swift-case-paths/commit/90e4e67ee18f72f2b1251211214ffc1fba60e540)):

```
name               time       std        iterations
---------------------------------------------------
Success.Manual      34.000 ns ± 146.77 %    1000000
Success.Reflection 151.000 ns ± 121.01 %    1000000
Failure.Manual      35.000 ns ± 231.80 %    1000000
Failure.Reflection  65.000 ns ± 191.00 %    1000000
```

New version:

```
name               time       std        iterations
---------------------------------------------------
Success.Manual      33.000 ns ± 226.73 %    1000000
Success.Reflection 128.000 ns ±  85.30 %    1000000
Failure.Manual      37.000 ns ± 149.48 %    1000000
Failure.Reflection  73.000 ns ± 140.36 %    1000000
```

mayoff:
Surprisingly, unwrapping just the cached tag first to reject non-matching values (instead of unwrapping the whole cached extractor) consistently saves a few seconds in the Failure.Reflection benchmark.

```
name               time       std        iterations
---------------------------------------------------
Success.Manual      34.000 ns ± 261.66 %    1000000
Success.Reflection 135.000 ns ± 116.75 %    1000000
Failure.Manual      36.000 ns ± 221.68 %    1000000
Failure.Reflection  66.000 ns ± 174.90 %    1000000
```

mayoff:
I think I'm done dinking this.

I've tried a few variations to knock nanoseconds off the Failure.Reflection benchmark without success, like storing -1 for cachedTag instead of using an optional, but was unable to make it consistently faster. I suspect it's consistently slower than the main branch by ~3 nanos because the closure captures two words (the cached tag and the strategy) instead of just one (the cached tag).

I thought the whole private helpers section of the file was kind of jumbled so I reorganized it. The code is all the same, but hopefully it's in a more logical order now so it should be easier to read from top to bottom.

iampatbrown:
Really awesome work @mayoff! Thanks heaps for the informative comments and commits :) it's looking pretty close to optimal.

stephencelis (approved):
We just read through and looks great, thanks @mayoff!

We're going to merge as-is to keep the record of your comments in Git. We'll prune them afterwards, as well as clean up the styling to be a lil more consistent with the rest of the repo.

PR #38, opened 2021-07-12
Stylistic cleanup for metadata extraction

stephencelis:
This does some stylistic cleanup from @mayoff's excellent PR, #37.

- Prefer implicit `return`
- Prefer explicit `self`
- Minimize comments. @mayoff's literate style was very helpful in getting an understanding of some of the code, and will live in the history of the project. We can consider bringing back context in the future if we find any of it helpful, but let's minimize comments for now.

@mayoff If you're cool with this, lemme know and I'll merge! A quick once-over to make sure I didn't break anything or make any incorrect assumptions would be helpful. All the tests appear to still pass.

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -87,16 +87,19 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
       return cachedStrategy.extract(from: root, tag: rootTag)
     }
 
-    guard let value = Strategy<Root, Value>(tag: rootTag).extract(from: root, tag: rootTag)
+    let rootStrategy = Strategy<Root, Value>(tag: rootTag)
+    guard let value = rootStrategy.extract(from: root, tag: rootTag)
     else { return nil }
 
     let embedTag = metadata.tag(of: embed(value))
     cachedTag = embedTag
-    cachedStrategy = Strategy<Root, Value>(tag: embedTag)
-
-    guard embedTag == rootTag else { return nil }
-
-    return value
+    if embedTag == rootTag {
+      cachedStrategy = rootStrategy
+      return value
+    } else {
+      cachedStrategy = Strategy<Root, Value>(tag: embedTag)
+      return nil
+    }
   }
 }
 
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -6,7 +6,7 @@ extension CasePath {
   /// - Parameter embed: An enum case initializer.
   /// - Returns: A case path that extracts associated values from enum cases.
   public static func `case`(_ embed: @escaping (Value) -> Root) -> CasePath {
-    return self.init(
+    self.init(
       embed: embed,
       extract: CasePaths.extract(embed)
     )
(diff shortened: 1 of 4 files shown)


Sources/CasePaths/EnumReflection.swift
```diff
@@ -67,89 +67,48 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
-  /// The case stores a protocol existential, either directly or indirectly. This strategy is used when the enum case's associated value type is a protocol existential and the `CasePath`'s `Value` is a type that conforms to the protocol (but is not itself the protocol existential).
-  case existential(get: (Enum) -> Any?)
+  case unimplemented
+  case void
 }
 
 extension Strategy {
-  /// Choose the appropriate `Strategy` to extract a `Value` from an `Enum` if that `Enum`'s case tag is `tag`. If `assumedAssociatedValueType` is nil, I'll look up the associated value type in the `Enum` metadata.
-  init?(tag: UInt32, assumedAssociatedValueType: Any.Type? = nil) {
+  init(tag: UInt32, assumedAssociatedValueType: Any.Type? = nil) {
```
stephencelis, line 111: I made this unfailable, since it seems like we can always return `.unimplemented` when it fails.

Sources/swift-case-paths-benchmark/main.swift
```diff
@@ -26,6 +26,10 @@ let success = BenchmarkSuite(name: "Success") {
   $0.benchmark("Reflection") {
     precondition(reflection.extract(from: enumCase) == 42)
   }
+
+  $0.benchmark("Reflection (uncached)") {
+    precondition((/Enum.associatedValue).extract(from: enumCase) == 42)
```
stephencelis, line 31: Felt like we should also have an understanding of the baseline when working with case paths that aren't cached in a larger system, _e.g._ a one-off `compactMap(/Result.success)`. It's still quite fast!

```
name                          time       std        iterations
--------------------------------------------------------------
Success.Manual                 32.000 ns ± 679.72 %    1000000
Success.Reflection            126.000 ns ± 391.25 %    1000000
Success.Reflection (uncached) 814.000 ns ± 131.98 %    1000000
Failure.Manual                 35.000 ns ± 767.22 %    1000000
Failure.Reflection             70.000 ns ± 519.43 %    1000000
Failure.Reflection (uncached) 552.000 ns ± 172.81 %    1000000
```

Sources/CasePaths/EnumReflection.swift
mayoff: If it's no longer going to print a warning when `init` fails, then I see no reason for `init` to be failable.

Sources/CasePaths/EnumReflection.swift
```diff
@@ -67,89 +67,48 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
-  guard metadata.typeDescriptor.fieldDescriptor != nil else {
-    #if DEBUG
-      print(
-        "\(#function) can never extract values from \(Root.self) because the metadata for \(Root.self) is incomplete!"
-      )
-    #endif
-    return { _ in nil }
-  }
+  var cachedTag: UInt32?
+  var cachedStrategy: Strategy<Root, Value>?
```
mayoff, line 79: I had `cachedStrategy` non-optional because it was maybe ~2ns faster. 🤷

Sources/CasePaths/EnumReflection.swift
```diff
@@ -67,89 +67,48 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
-      return nil
-    }
+    guard let value = Strategy<Root, Value>(tag: rootTag).extract(from: root, tag: rootTag)
+    else { return nil }
 
-    let embeddedValue = embed(value)
-    let embedTag = metadata.tag(of: embeddedValue)
+    let embedTag = metadata.tag(of: embed(value))
     cachedTag = embedTag
+    cachedStrategy = Strategy<Root, Value>(tag: embedTag)
```
mayoff, line 95: My assumption was that `swift_getTypeByMangledNameInContext` is slow compared to everything else this function does, so I wanted to avoid redundantly creating the same `Strategy` twice. I'm sure it's a trivial savings if the `CasePath` object is used repeatedly.

Sources/CasePaths/EnumReflection.swift
```diff
@@ -67,89 +67,48 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
...
-        cachedStrategy = embedStrategy
-      } else {
-        #if DEBUG
-          print(
-            "\(#function) can never extract values from a \(Root.self) case like \(embeddedValue) because I don't know how to convert its associated value to \(Value.self)!"
-          )
-        #endif
-        cachedStrategy = .unimplemented
-      }
+    guard cachedTag == rootTag else { return nil }
```
mayoff, line 97: This should be `embedTag == rootTag`. And it was, before I started working on the code. I messed it up.

mayoff (commented):
looks good to me

Sources/CasePaths/EnumReflection.swift
stephencelis: Ha. I think it’s ok to not micro-optimize further. This spelling made a little more sense to me since `nil` means uncached.

Sources/CasePaths/EnumReflection.swift
stephencelis: This should still only get called once, right? As soon as it’s cached it’ll always bail out early? Maybe I’m missing something.

mayoff: You already created a strategy to extract `root`s value. If `root` had the correct tag, then you could store that strategy in `cachedStrategy` instead of creating it anew, saving one name-to-type lookup. Not a big deal as long as `CasePath`s are reused.

stephencelis: Ah yeah, good call! Lost track of where `rootStrategy` was used during the refactor.

mbrandonw (approved):

PR #41, opened 2021-07-19
fix issue #40 (case initializer promoted to return Optional)

### LINKED ISSUE #40: Extraction from optional root
Hi,
updating from `0.2.0` to `0.5.0` resulted in an unexpected behavior for me. It seems that is no longer possible to extract a CasePath from from an optional value. 
The following test case works fine in `0.2.0`, but fails on `0.5.0`:

```
  func testExtractFromOptionalRoot() {
    enum Authentication {
      case authenticated(token: String)
      case unauthenticated
    }
    
    // Fails
    let optionalAuthentication: Authentication? = .authenticated(token: "deadbeef")
    XCTAssertNotNil(extract(case: Authentication.authenticated, from: optionalAuthentication))
  }
```

It has probably something to do with moving away from `Mirror`, but i didn't dig deeper.

Would be good to know if you are aware of this change in the behavior and if it's intended.

mayoff:
Swift promotes (A) -> B to (A) -> B? if necessary to make a statement
type-check. This can lead to an enum case initializer being promoted and
making extract erroneously return nil. Fixing it is hard, so instead I'm
detecting the problem and asserting.

Fixes #40.

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -321,7 +321,7 @@ extension EnumMetadata {
   }
 
   func isOptional() -> Bool {
-    // All Optional types shared a common EnumTypeDescriptor.
+    // All Optional types share a common EnumTypeDescriptor.
     let optionalTypeDescriptor = EnumMetadata(assumingEnum: Void?.self).typeDescriptor
     return typeDescriptor == optionalTypeDescriptor
   }
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -162,11 +162,6 @@ extension Strategy {
       }
       self.init(tag: tag, assumedAssociatedValueType: Value.self)
 
-    } else if metadata.wrappedTypeIfOptional() == avType {
-      // Enum == Optional<Inner>
-      //
-      self = .unimplemented
-
     } else if ExistentialMetadata(avType) != nil {
       // Convert protocol existentials to `Any` so that they can be cast (`as? Value`).
       let anyStrategy = Strategy<Enum, Any>(nonExistentialTag: tag)
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -11,6 +11,13 @@ extension CasePath {
       extract: CasePaths.extract(embed)
     )
   }
+
+  public static func optionalCase<Wrapped>(_ embed: @escaping (Value) -> Wrapped?) -> CasePath where Root == Wrapped? {
+    self.init(
+      embed: embed,
+      extract: CasePaths.extract(embed)
+    )
+  }
 }
 
 extension CasePath where Value == Void {
@@ -50,6 +57,10 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
   CasePaths.extract(embed)(root)
 }
 
+public func extract<Root, Value>(case embed: @escaping (Value) -> Root?, from root: Root?) -> Value? {
+  CasePaths.extract(embed)(root)
+}
+
 /// Returns a function that can attempt to extract associated values from the given enum case
 /// initializer.
 ///
@@ -66,7 +77,7 @@ public func extract<Root, Value>(case embed: @escaping (Value) -> Root, from roo
 ///   otherwise undefined.
 /// - Parameter embed: An enum case initializer.
 /// - Returns: A function that can attempt to extract associated values from an enum.
-public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -> (Value?) {
+public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -> Value? {
   guard
     let metadata = EnumMetadata(Root.self),
     metadata.typeDescriptor.fieldDescriptor != nil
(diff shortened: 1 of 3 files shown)


stephencelis:
@mayoff I took a quick stab at fixing the problem here: https://github.com/pointfreeco/swift-case-paths/compare/fix-optional-promotion

I'm sure it could be cleaned up (and you may have suggestions on how to do so), but do you see anything wrong with this approach?

mayoff:
@stephencelis Unfortunately, your approach won't work. I've added a failing test case for it: https://github.com/mayoff/swift-case-paths/commit/7cb4382c5c06b9edac31350a93e252220bf3d0fd

I added the test case to my fix and it passes.

My fix has some code duplication because I needed to overload `extract` to catch the `Optional` case and, in the inner closure, unwrap optionals in two places. I didn't see a pleasant way to extract the common bits.

mayoff:
To clarify why your solution doesn't work:

In my test case, both `E` (the enum type) and `E?` are stored in a single word:

- `E?.none` is stored as zero. Zero is not a valid `E`.
- `E?.some(.c1(o))` is stored as the pointer to `o`.
- `E?.some(.c2(o))` is stored as the pointer to `o`, with its high bit set.

Your fix uses the `E?` value witness table to project an `E?` to an `E` and then blindly treats the `E` as a `TestObject` pointer. Since you don't use the `E` value witness table to project the `E` to a `TestObject`, you don't strip the tag bit from the `E.c2` case and get an invalid `TestObject` pointer. (This need to double-project is why I initially decided this was too hard to bother fixing!)

Sources/CasePaths/EnumReflection.swift
```diff
@@ -75,6 +86,16 @@ public func extract<Root, Value>(_ embed: @escaping (Value) -> Root) -> (Root) -
     return { _ in nil }
   }
 
+  // https://github.com/pointfreeco/swift-case-paths/issues/40
+  if
+    let wrappedType = metadata.wrappedTypeIfOptional(),
+    wrappedType != Value.self,
+    EnumMetadata(wrappedType) != nil
+  {
+    assertionFailure("embed parameter must not be promoted to return Optional (https://github.com/pointfreeco/swift-case-paths/issues/40)")
```
stephencelis, line 95: Is it still possible to hit this code path?

mayoff: Seems unlikely to be hit without heroic effort.

stephencelis:
> Your fix uses the `E?` value witness table to project an `E?` to an `E` (which is actually a no-op) and then blindly treats the `E` as a `TestObject` pointer. Since you don't use the `E` value witness table to project the `E` to a `TestObject`, you don't strip the tag bit from the `E.c2` case and get an invalid `TestObject` pointer. (This need to double-project is why I initially decided this was too hard to bother fixing!)

I was afraid of that. Thanks for prepping the failing case!

Sources/CasePaths/EnumReflection.swift
```diff
@@ -11,6 +11,13 @@ extension CasePath {
       extract: CasePaths.extract(embed)
     )
   }
+
+  public static func optionalCase<Wrapped>(_ embed: @escaping (Value) -> Wrapped?) -> CasePath where Root == Wrapped? {
```
stephencelis, line 15: Can this be an overload, as well?
```suggestion
  public static func `case`<Wrapped>(_ embed: @escaping (Value) -> Wrapped?) -> CasePath where Root == Wrapped? {
```

Sources/CasePaths/EnumReflection.swift
stephencelis: I wonder if we should maybe not worry about maintaining it and the associated runtime code, then. What do you think?

Sources/CasePaths/Operators.swift
```diff
@@ -26,6 +26,12 @@ public prefix func / <Root, Value>(
   .case(embed)
 }
 
+public prefix func / <Root, Value>(
+  embed: @escaping (Value) -> Root?
+) -> CasePath<Root?, Value> {
+  .optionalCase(embed)
```
stephencelis, line 32: (Would have to rename this:)
```suggestion
  .case(embed)
```

stephencelis:
Brandon and I discussed and think we should support this, even though overloads can cause the compiler to churn in some cases.

One other place that may need an overload is where we treat case path syntax as extract functions, e.g.:

```swift
results.compactMap(/Result.success)
```

See: https://github.com/pointfreeco/swift-case-paths/blob/173007350b9f5202fa7d9c2842407a363ca076a2/Sources/CasePaths/Operators.swift#L84-L89

It might be an unlikely code path, but if `results: [Result<Int, Never>?]`, then I think the above would fail at runtime? 

Hopefully this is the extent of overloads for now, and maybe we'll have a more focussed API in time for a 1.0.

Sources/CasePaths/EnumReflection.swift
mayoff: I tried that and it didn't work.

Sources/CasePaths/Operators.swift
mayoff: If it works for you, ok. I seem to recall when I tried overloading the name, it called the non-optional version somewhere that it should have called the optional version.

stephencelis: Just wondering how that would be possible since the module-level `extract` function and `/` operators seem to overload just fine. Is there something strange with the generic context? Can you cook up a failure that I can take a look at?

Sources/CasePaths/EnumReflection.swift
stephencelis: Could be we need to specify generics differently. If you can remember a case that didn't work I can mess with it.

Sources/CasePaths/Operators.swift
mayoff: I just double-checked. If you rename `optionalCase` to `case`, the `/` overload that currently calls `optionalCase` calls the wrong overload of `case`.

mayoff: ... and then it hits the “embed parameter must not be promoted” assertion in `extract`.

Sources/CasePaths/EnumReflection.swift
mayoff: I think I could cook up a scenario involving generics that would call the wrong overload... but I really don't want to. That's what I meant by heroic effort. 😅 I'd personally leave the assertion there, just in case.

Sources/CasePaths/EnumReflection.swift
mayoff: Rename it to `case` and change the `/` overload that calls it. In Xcode 13 beta 3, the `/` operator then calls the wrong `case` overload.

Sources/CasePaths/Operators.swift
stephencelis: Dang, yeah. For whatever reason it seems that optional promotion does its thing before overload resolution?

stephencelis:
@mayoff Thanks again for exploring this! We're going to merge as is and I'll make a few changes after, mainly I'm going to make `optionalCase` internal/private, and see if I can address the case path expression overload, as well.

I think now that CasePaths has matured we will probably eliminate a lot of the API surface area. The static `.case`s seem unnecessary. We introduced them because we though the prefix operator might scare some folks off, but we haven't seen anyone prefer them to `/` in the wild. And at that point, the top-level `extract` functions seem unnecessary when it's easy enough to use a case path expression or case path. We'll look into deprecating these soon, but if you (or anyone) has any concerns with that direction, let us know!

PR #55, opened 2021-08-10
handle the Error existential properly (fixes #53)

### LINKED ISSUE #53: EXC_BAD_ACCESS when extracting a `Result.success` from a failure, if the result is `Result<_, Error>`
This happens in 0.7.0, I haven't tried other releases.

I've encountered this crash in the wild, which I had to work around by not using case paths for any `Result` where the `Failure` generic type is `Error`. If you try to add this unit test to the test suite in this repo, you can reproduce it:

```swift
  func testExtractSuccessFromFailedResultWithErrorProtocolError() throws {
    struct MyError: Error {}
    let x = Result<String, Error>.failure(MyError()) 

    // since it's `Error`, execution traps at the following line with an EXC_BAD_ACCESS. 
    // If the  result type was over `MyError`, it would work correctly.
    XCTAssertNil((/Result.success).extract(from: x))
  }
```

There must be some swift runtime issue with Error protocol types vs concrete types that conform to `Error`, since it works if the type is `Result<String, MyError>` instead. I'm not too well versed in the swift runtime details to know how to fix this, except to work around it by not using case paths for these kinds of `Result` types for now.

If the value was a `.success` instead of a `.failure`, extraction does not crash, only when trying to extract a `.success` from this specific kind of Result if the value is a `.failure`. Extracting a `.failure` from this value also seems to work as expected.

mayoff:
Fixes #53.

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -282,6 +282,16 @@ extension Strategy {
       self.init(tag: tag, assumedAssociatedValueType: Value.self)
 
     } else if ExistentialMetadata(avType) != nil {
+#if _runtime(_ObjC)
+        if avType == Error.self {
+          // https://github.com/pointfreeco/swift-case-paths/issues/53
+          // For Objective-C interop, the Error existential is a pointer to an NSError-compatible (and thus AnyObject-compatible) object.
+          let errorStrategy = Strategy<Enum, AnyObject>(nonExistentialTag: tag)
+          self = .existential { errorStrategy.extract(from: $0, tag: tag) }
+          return
+        }
+#endif
+
       // Convert protocol existentials to `Any` so that they can be cast (`as? Value`).
       let anyStrategy = Strategy<Enum, Any>(nonExistentialTag: tag)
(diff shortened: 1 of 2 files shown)


mayoff:
Oh dear... I don't have a Linux setup for compiling Swift.

Maybe try changing the `#if _runtime(_ObjC)` to `#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)`.

I'll push a new version if you want, but if you can just log in to a Linux system and try it out, that might save some round-trips through CI.

stephencelis:
Thanks for submitting! I'm not going to be around a computer, but can try tomorrow, but we don't mind if you keep pushing to troubleshoot! The builds should be fast.

If you have Docker installed (via Homebrew or Docker for Mac) you can also run `make test-linux` and it should take care of everything for you.

mbrandonw:
From my testing it seems that the code inside the `#if` is needed for Linux as well as Darwin: 1fc81ea

mbrandonw (approved):
Seems good to me, but Stephen and I will discuss more tomorrow.

Thanks for the help @mayoff!

PR #64, opened 2021-12-27
Introduce IfCaseLetEmbed

maximkrouk:
It enables you to implement enum properties easier

```swift
enum Enum {
  case a(Int)
  case b(Bool)
  
  var a: Int? {
    get { (/Self.a).extract(from: self) }
    set { (/Self.a).ifCaseLetEmbed(newValue, in: &self) }
  }
  
  var b: Bool? {
    get { (/Self.b).extract(from: self) }
    set { (/Self.b).ifCaseLetEmbed(newValue, in: &self) }
  }
}
```

instead of 

```swift
enum Enum {
  case a(Int)
  case b(Bool)
  
  var a: Int? {
    get { (/Self.a).extract(from: self) }
    set {
      guard case .a = self else { return }
      self = .a(newValue)
    }
  }
  
  var b: Bool? {
    get { (/Self.b).extract(from: self) }
    set {
      guard case .b = self else { return }
      self = .b(newValue)
    }
  }
}
```

### DIFF
--- a/Sources/CasePaths/CasePaths.swift
+++ b/Sources/CasePaths/CasePaths.swift
@@ -54,3 +54,29 @@ extension CasePath where Value: LosslessStringConvertible, Root == String {
     )
   }
 }
+
+extension CasePath {
+  /// Helper function for enum properies
+  ///
+  /// Usage
+  /// ```
+  /// enum Enum {
+  ///   case a(Int)
+  ///   case b(Bool)
+  ///
+  ///   var a: Int? {
+  ///     get { (/Self.a).extract(from: self) }
+  ///     set { (/Self.a).ifCaseLetEmbed(newValue, in: &self) }
+  ///   }
+  ///
+  ///   var b: Bool? {
+  ///     get { (/Self.b).extract(from: self) }
+  ///     set { (/Self.b).ifCaseLetEmbed(newValue, in: &self) }
+  ///   }
+  /// }
+  /// ```
+  public func ifCaseLetEmbed(_ value: Value?, in root: inout Root) {
+    guard extract(from: root) != nil, let value = value else { return }
+    root = embed(value)
+  }
+}


stephencelis:
@maximkrouk Have you seen [`CasePath.modify`](https://github.com/pointfreeco/swift-case-paths/blob/de7bddf2254d88d3c2d7896b1fbb2d1e694d24dd/Sources/CasePaths/CasePath.swift#L35-L50)? I think you could use it to achieve something similarly succinct:

```swift
enum Enum {
  case a(Int)
  case b(Bool)
  
  var a: Int? {
    get { (/Self.a).extract(from: self) }
    set { try? (/Self.a).modify(&self) { $0 = newValue } }
  }
  
  var b: Bool? {
    get { (/Self.b).extract(from: self) }
    set { try? (/Self.b).modify(&self) { $0 = newValue } }
  }
}
```

iampatbrown:
I was just about to post the exact same example :P 
Was also going to mention to be cautious of someting like

```swift
var state = Enum.a(1)
state.a = nil
state.a == 1 // true
```

maximkrouk:
@stephencelis, `modify` makes it too verbose, it's even throwable 🌚
IfCaseLet will provide really tiny setters a call site, very similar to getters of enum properties.

In general I'm kinda okay with
```swift
guard let value = newValue, case .case = self else { return }
self = .case(value)
```
and anyway I'm free to use IfCaseLet by creating an extension outside of the package, but I found it super convenient, so I decided to make a PR 🙂

maximkrouk:
@iampatbrown, yep, forgot this in the example above, but not at the IfCaseLet implementation 😎

I think the behavior is clear enough  (anyway maybe it worth mentioning in inline doc), the primary use case for it is enum properties and the behavior partially defined by it, also the case you trying to modify has a concrete type, providing an optional have no reason so should be ignored. I feel like IfCaseLet prefix tells that the value will be unwrapped, but if it's just a personal feeling then it might be worth adding a label like
`(Self.a).ifCaseLetEmbed(unwrapped: newValue, into: &self)`
But I'd prefer a shorter version 🙂

maximkrouk:
https://github.com/pointfreeco/swift-case-paths/pull/64#issuecomment-1001802806 is similar to https://github.com/pointfreeco/swift-case-paths/pull/62#issuecomment-998024589, but I still think that while we are unable to use different types for getter and a setter - function is less misleading than subscript and clear enough

But actually, I'm not sure that it's too misleading with a subscript too especially due to explicit protocol declaration, and I like the subscript API so I'd probably like both PRs to be merged 🌚

stephencelis:
> Edit: @stephencelis could probably do `try? (/Self.b).modify(&self) { $0 = newValue ?? $0 }` as well but I think using `if let newValue = newValue` is clearer

Nice, @iampatbrown! Since this is all just boilerplate I think shortening it further is fine, even if it's a little less clear :smile:

> @stephencelis, `modify` makes it too verbose, it's even throwable 🌚
> IfCaseLet will provide really tiny setters to a call site, very similar to getters of enum properties.

It may seem extreme, but one of the reasons we deprecated swift-enum-properties in favor of swift-case-paths was to eliminate this `nil`-writable surface area that @iampatbrown drew attention to, and that I brought up in #62. While it's easy enough to reason that writing `nil` is a no-op, this is code that should never happen, and it is possible to write code that _does_ write `nil` values at runtime, which is probably a bug.

Because of this, you may even prefer to write enum properties (and your `ifCaseLetEmbed` helper) with `precondition`s/`fatalError`s:

```swift
var a: Int? {
  get { (/Self.a).extract(from: self) }
  set {
    precondition(newValue != nil, "new value must not be nil")
    guard case .a = self else { return }
    self = .a(newValue!)
  }
}
```

And this is why we're a bit hesitant to introduce APIs that allow for `nil`-writability, since it introduces invalid runtime states that may need preconditions. Ideally Swift will evolve to introduce properties that allow for optional-chaining semantics (something like `optional get` or `nonoptional set`), or even better, introduce case paths directly to the language...though it may be awhile before an engineer with compiler experience takes on such a project :sweat_smile:

stephencelis:
@maximkrouk OK to close this out for the reasons we brought up earlier? Luckily you could release this API as a companion library, and if you do you should share it in the GitHub discussions. I think we want to avoid introducing an API that has to have an opinion on optional writability in the main library, though.

maximkrouk:
Well, I close this now, maybe one day we'll get `nonoptional set`s 🙏😅

PR #65, opened 2021-12-31
Added check for class constrained existential

### LINKED ISSUE #63: EXC_BAD_ACCESS extracting existensial
This was partially solved in https://github.com/pointfreeco/swift-case-paths/issues/53 but it seems that the fix is needed for all object-bound existential types.

Here's a rudimentary case that reproduces the crash (tested with Xcode 13.1, iOS 13.5 simulator):
```swift
private protocol TestObjectProtocol: AnyObject {}

func testObjectBoundExistensialPayload() {
  class Class: TestObjectProtocol {}
  enum Enum {
    case int(Int)
    case objectBoundExistensial(TestObjectProtocol)
  }
  let intPath = /Enum.int
  let object = Class()
  for _ in 1...2 {
      XCTAssertNil(intPath.extract(from: .objectBoundExistensial(object)))
  }
}
```

iampatbrown:
Possible option to fix #63

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -281,8 +281,8 @@ extension Strategy {
       }
       self.init(tag: tag, assumedAssociatedValueType: Value.self)
 
-    } else if ExistentialMetadata(avType) != nil {
-      if avType == Error.self {
+    } else if let avMetadata = ExistentialMetadata(avType) {
+      if avType == Error.self || avMetadata.isClassConstrained {
         // For Objective-C interop, the Error existential is a pointer to an NSError-compatible
         // (and thus AnyObject-compatible) object.
         let strategy = Strategy<Enum, AnyObject>(nonExistentialTag: tag)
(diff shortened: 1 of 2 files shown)


Sources/CasePaths/EnumReflection.swift
```diff
@@ -281,8 +281,8 @@ extension Strategy {
       }
       self.init(tag: tag, assumedAssociatedValueType: Value.self)
 
-    } else if ExistentialMetadata(avType) != nil {
-      if avType == Error.self {
+    } else if let avMetadata = ExistentialMetadata(avType) {
+      if avType == Error.self || avMetadata.isClassConstrained {
         // For Objective-C interop, the Error existential is a pointer to an NSError-compatible
```
iampatbrown, line 286: probably need to update this comment as well.

stephencelis (approved):
Great find, @iampatbrown!

Tests/CasePathsTests/CasePathsTests.swift
```diff
@@ -359,6 +361,26 @@ final class CasePathsTests: XCTestCase {
     }
   }
 
+  func testClassConstrainedExistensial() {
```
tgrapperon, line 364:
```suggestion
  func testClassConstrainedExistential() {
```
Little tipo

iampatbrown: Thanks @tgrapperon. Good catch :)

PR #66, opened 2022-02-12
Combine operator: extract

jtouzy:
Hi there!

I was wondering if you could be interested to integrate this Combine operator in the case-paths library.
We use it for a while now, and it's pretty helpful for making simple stream chains.

As an example in our usage :

```swift
imageService
      .loadImage(request)
      .filter(\.isSuccess)
      .extract(/AnyImage.FetchStatus.success)
      .eraseToAnyPublisher()
```

PS: I didn't implement any tests for now, it's just to see the interest : maybe you don't want any Combine code in this repository, but I think it's very complimentary with the library for Combine users.

### DIFF
--- a/Sources/CasePaths/Combine.swift
+++ b/Sources/CasePaths/Combine.swift
@@ -1,0 +1,43 @@
+#if canImport(Combine)
+import Combine
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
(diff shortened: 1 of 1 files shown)


jtouzy:
Never thought about using a compactMap mixed to the operator... good point !
Just tried and get the same behavior. I'm closing this 😬 Thanks for the tip !

Issue #67, opened 2022-04-19
`..` operator associativity


nonameplum:
Hey,

I wonder why the `CasePathCompositionPrecedence` precedence group has configured `right` `associativity`.

Especially the shortcoming is visible to me in such case:

```swift
enum Case {
    case one(One)
    case none

    enum One {
        case two(Two)
        case none

        enum Two {
            case value(String)
            case none
        }
    }
}

let root = /Case.one..Case.One.two..Case.One.Two.value // error: cannot convert value of type '(Case.One.Two) -> Case.One' to expected argument type 'CasePath<Case.One, Case.One.Two>'

// brakets required
let root = (/Case.one..Case.One.two)..Case.One.Two.value // this compiles
```

If the `associativity` would be configured to be from left to right by:

```swift
precedencegroup CasePathCompositionPrecedence {
  associativity: left
}
```

then

```swift
let root = /Case.one..Case.One.two..Case.One.Two.value
```

compiles just fine.

stephencelis:
Hi @nonameplum,

We probably hadn't noticed this due to a preference for the `appending(path:)` method. Would you like to submit a PR with a compiling test?

PR #69, opened 2022-04-26
Issue #67 Change CasePathCompositionPrecedence associativity

nonameplum:
Change `associativity` from `right` to `left`, which improves the ergonomics of the `..` operator in the case of the nested enums.

This addresses issue #67.


stephencelis (approved):
Looks good! Thanks for PRing!

PR #78, opened 2022-06-08
Add SPI DocC support

stephencelis:


Package.swift
```diff
@@ -30,3 +30,10 @@ let package = Package(
     ),
   ]
 )
+
+#if swift(>=5.6)
+  // Add the documentation compiler plugin if possible
+  package.dependencies.append(
+    .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0")
+  )
+#endif
```
stephencelis, line 39: This step is technically not necessary, but kinda nice to have the plug-in in Xcode 14.

mbrandonw (approved):

PR #90, opened 2022-10-13
Make CasePath sendable

### LINKED ISSUE #89: Sendable Case Paths?
I've been getting some warnings when I capture a 'non-sendable' case path in a `@Sendable` closure.

It there any reason `CasePath` can't be marked `Sendable` where the `Root` is already `Sendable`? I _think_ this will be okay, but would like to make sure

🤔

mbrandonw:
We can make `CasePath` sendable by locking around the part where we memoize some data from the runtime metadata. There's a chance we missed other parts that need locking, so feel free to let us know if you spot something!

Fixes #89

### DIFF
--- a/Sources/CasePaths/CasePath.swift
+++ b/Sources/CasePaths/CasePath.swift
@@ -78,7 +78,7 @@ public struct CasePath<Root, Value> {
   }
 }
 
-#if canImport(_Concurrency)
+#if canImport(_Concurrency) && compiler(>=5.5.2)
   extension CasePath: @unchecked Sendable {}
 #endif
 
--- a/Sources/CasePaths/CasePath.swift
+++ b/Sources/CasePaths/CasePath.swift
@@ -2,7 +2,7 @@
 /// value.
 ///
 /// This type defines key path-like semantics for enum cases.
-public struct CasePath<Root, Value> {
+public struct CasePath<Root, Value>: @unchecked Sendable {
   private let _embed: (Value) -> Root
   private let _extract: (Root) -> Value?
 
--- a/Sources/CasePaths/CasePath.swift
+++ b/Sources/CasePaths/CasePath.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// A path that supports embedding a value in a root and attempting to extract a root's embedded
 /// value.
 ///
@@ -11,9 +13,20 @@ public struct CasePath<Root, Value> {
   /// - Parameters:
   ///   - embed: A function that always succeeds in embedding a value in a root.
   ///   - extract: A function that can optionally fail in extracting a value from a root.
-  public init(embed: @escaping (Value) -> Root, extract: @escaping (Root) -> Value?) {
-    self._embed = embed
-    self._extract = extract
+  public init(
+    embed: @escaping (Value) -> Root,
+    extract: @escaping (Root) -> Value?
+  ) {
+    self._embed = {
+      lock.lock()
+      defer { lock.unlock() }
+      return embed($0)
+    }
+    self._extract = {
+      lock.lock()
+      defer { lock.unlock() }
+      return extract($0)
+    }
   }
 
   /// Returns a root by embedding a value.
@@ -76,3 +89,5 @@ extension CasePath: CustomStringConvertible {
 }
 
 struct ExtractionFailed: Error {}
+
+private let lock = NSRecursiveLock()
--- a/Sources/CasePaths/CasePath.swift
+++ b/Sources/CasePaths/CasePath.swift
@@ -78,7 +78,7 @@ public struct CasePath<Root, Value> {
   }
 }
 
-#if swift(>=5.5)
+#if canImport(_Concurrency)
   extension CasePath: @unchecked Sendable {}
 #endif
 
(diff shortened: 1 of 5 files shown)


iampatbrown:
@mbrandonw do you think the `init` should require sendable closures? Or do you think protecting against this use case is too restrictive?

mbrandonw:
@iampatbrown My thought is that all access to embed is guarded by a lock, and so it's ok to use non-`@Sendable` closures and even non-`Sendable` roots and values.

This is how Swift's key path works too. A key path `\Root.value` is sendable regardless of the sendability of `Root` and `Value`.

Now, that is my understanding of the situation. I could be wrong! Would love to know if there's something I'm missing and if we need to be more restrictive. Perhaps I will write a test for a "bad" case path and see if I can get something bad to happen.

iampatbrown:
@mbrandonw my suspicion was that it could be abused by passing a non sendable closure to the init. Possibly something that is reading an external mutable value. 
I haven't had time to test this out and think the use case would be pretty rare and maybe even completely incorrect :P
I have used this technique before for extracting a selected value from a list but probably wouldn't do it these days.

mbrandonw:
@iampatbrown I think you are correct. I was able to create a "bad" case path to mutated some shared storage and was able to produce a data race in a test.

I think we have fixed that now though. We now completely lock all access to embed and extract when constructing a case path. Can you think of any more weird things one could do to break this?

tgrapperon:
Without weighting in the PR, here are the benchmark results between `main` and this PR:
```
Current:
name                          time       std        iterations
--------------------------------------------------------------
Success.Manual                 41.000 ns ± 112.18 %    1000000
Success.Reflection             83.000 ns ± 106.03 %    1000000
Success.Reflection (uncached) 334.000 ns ±  42.44 %    1000000
Failure.Manual                 41.000 ns ± 113.15 %    1000000
Failure.Reflection             83.000 ns ±  67.90 %    1000000
Failure.Reflection (uncached) 333.000 ns ±  41.24 %    1000000

Sendable:
name                          time       std        iterations
--------------------------------------------------------------
Success.Manual                 42.000 ns ± 226.83 %    1000000
Success.Reflection            125.000 ns ± 111.91 %    1000000
Success.Reflection (uncached) 458.000 ns ±  63.48 %    1000000
Failure.Manual                 42.000 ns ± 186.97 %    1000000
Failure.Reflection             83.000 ns ± 121.09 %    1000000
Failure.Reflection (uncached) 417.000 ns ±  65.96 %    1000000
```
So that's a 50% increase on successful matches, but we're in the sub µs domain, so I don't know how bad it effectively is.

iampatbrown:
@mbrandonw oooo nice fix! That was the main one. If I think of anything else, I'll let you know.

mbrandonw:
@tgrapperon yeah I think the changes in benchmarks are probably small enough to be worth it.

FWIW, here is how the benchmarks fare on my computer, which is slightly different from yours:

```
Sendable
--------
Success.Manual                 42.000 ns ± 210.71 %    1000000
Success.Reflection             84.000 ns ± 100.28 %    1000000
Success.Reflection (uncached) 375.000 ns ±  69.05 %    1000000
Failure.Manual                 42.000 ns ± 170.72 %    1000000
Failure.Reflection             83.000 ns ±  80.62 %    1000000
Failure.Reflection (uncached) 375.000 ns ±  56.29 %    1000000

main
----
Success.Manual                 41.000 ns ±  81.32 %    1000000
Success.Reflection             83.000 ns ±  54.26 %    1000000
Success.Reflection (uncached) 333.000 ns ±  21.11 %    1000000
Failure.Manual                 41.000 ns ± 128.27 %    1000000
Failure.Reflection             42.000 ns ± 126.52 %    1000000
Failure.Reflection (uncached) 292.000 ns ±  24.71 %    1000000

```

tgrapperon:
@mbrandonw Ah, I made the mistake of running them on battery, which lower the performance. If I'm plugging it, I'm getting the same results as you are, M1 Pro on 14.1b3.
This is very likely negligible in real-life apps.

Sources/CasePaths/CasePath.swift
```diff
@@ -78,7 +78,7 @@ public struct CasePath<Root, Value> {
   }
 }
 
-#if swift(>=5.5)
+#if canImport(_Concurrency)
```
tgrapperon, line 81: We don't need `#if canImport(_Concurrency) && compiler(>=5.5.2)` like one previously did in TCA, right?

mbrandonw: oh maybe that's the magical incantation we need to get this finally passing on CI 😤

stephencelis (approved):

mjburghard:
@mbrandonw Can you explain why the lock is a global instead of an instance variable? The answer may be very obvious, but I am not yet familiar with all aspects of Swift Concurrency.

mbrandonw:
Hi @mjburghard, this is done because when a case path is constructed with only an embed (which means it uses the reflection machinery), as is the case when doing `/Enum.case`, then there is a bit of mutable state managed in the library that needs to be synchronized. And locally locking is not enough.

We even have a test that catches this. If you refactor `CasePath` to hold onto a lock instead of using the global lock, you will get a test failure.

Issue #71, opened 2022-05-08
Fail to extract cases with 1-case enums and `Any` associated values


tgrapperon:
I've extracted from TCA's [#1086](https://github.com/pointfreeco/swift-composable-architecture/issues/1086) the following minimal suite that doesn't pass when built in `Release` mode:
```swift
import CasePaths
import XCTest

public enum RootAction: Equatable {
  case branch(BranchAction)
//  case noop // Add this and extraction succeeds
}

public enum BranchAction: Equatable {
  case firstAction(Bool)
  case secondAction(Bool)
  case leaf(LeafAction) // Remove this and extraction succeeds
}

public enum LeafAction: Equatable {
  case value(Value) // Or case value(Any) with explicit `Equatable` conformance.
}

public struct Value: Equatable {
  var value: Any
  public static func == (lhs: Value, rhs: Value) -> Bool { true }
}

class CasePathIssueTests: XCTestCase {
  func testFirst() throws {
    XCTAssertEqual(
      (/RootAction.branch).extract(from: .branch(.firstAction(true))),
      .firstAction(true)
    )
  }
  func testSecond() throws {
    XCTAssertEqual(
      (/RootAction.branch).extract(from: .branch(.secondAction(true))),
      .secondAction(true)
    )
  }
}
```
A few comments:
- It passes in `Debug` mode
- If `.firstAction` has no associated value, it becomes the one that fails to extract, and `second` passes.

I'm not knowledgeable enough with runtime and the intricacies of the extraction algorithm to fix this issue.

stephencelis:
> As a fair amount of the code of this library is unsafe, and at the light of this issue, I wonder if it wouldn't be better to defensively test optimized builds in CI. If we implement a test case for this specific issue, it will pass under the radar with the current configuration, isn't it?

We'd be down to run tests in both DEBUG and RELEASE on CI. Should be something we can configure in the `.swiftpm` Xcode config files, or we could call out to `swift test` in the `Makefile`.

bbethke:
Hi! I recently encountered this bug as well. Are there any updates on the status of this issue?

Thanks for the awesome open-source libraries!

stephencelis:
@bbethke We unfortunately haven't been able to track down the right Swift runtime code for this yet 😞

Is the workaround mentioned above sufficient in the meantime?

bbethke:
Thanks for the update! Yes, that work around seems to do the trick.

tgrapperon:
[...]

PR #94, opened 2022-10-25
Workaround for runtime returning wrong enum tag

stephencelis:
Fixes #72.

In release builds, Swift <=5.7 may return the wrong tag for single-case enums that nest another enum. We can work around this by always returning "0" as the tag for single case enums.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test-swift:
  		--enable-test-discovery \
 		--parallel
 	swift test \
-	  -c release \
+		-c release \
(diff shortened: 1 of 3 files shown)


stephencelis:
Reported upstream in https://github.com/apple/swift/issues/61708

tgrapperon:
For future reference, the issue that was fixed is #71 and not #72.

stephencelis:
@tgrapperon Whoops! Typo. Fixed thanks!

PR #95, opened 2022-12-30
Add XCTUnwrap(_:case:) test helper.

mbrandonw:
This helper is analogous to the `XCTUnwrap` that comes with XCTest, but tuned specifically for enums and cases.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     strategy:
       matrix:
         xcode:
-          - 12.5.1
           - '13.0'
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     strategy:
       matrix:
         xcode:
-          - '13.4.1'
           - '14.2'
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,24 +29,6 @@ jobs:
       - name: Run tests
         run: make test-swift
 
-  macos-10-15:
-    name: MacOS 10.15
-    runs-on: macos-10.15
-    strategy:
-      matrix:
-        xcode:
-          - 11.3
-          - 11.7
-          - 12.4
-    steps:
-      - uses: actions/checkout@v2
-      - name: Select Xcode ${{ matrix.xcode }}
-        run: sudo xcode-select -s /Applications/Xcode_${{ matrix.xcode }}.app
-      - name: Print Swift version
-        run: swift --version
-      - name: Run tests
-        run: make test-swift
-
   linux:
     name: Ubuntu
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,12 +14,13 @@ concurrency:
 
 jobs:
   macos-11:
-    name: MacOS 11
-    runs-on: macos-11
+    name: MacOS 12
+    runs-on: macos-12
     strategy:
       matrix:
         xcode:
-          - '13.0'
+          - '13.4.1'
+          - '14.2'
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -1,21 +1,32 @@
...
     .library(
       name: "CasePaths",
       targets: ["CasePaths"]
     )
   ],
   dependencies: [
-    .package(name: "Benchmark", url: "https://github.com/google/swift-benchmark", from: "0.1.0")
+    .package(name: "Benchmark", url: "https://github.com/google/swift-benchmark", from: "0.1.0"),
+    //.package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay.git", from: "0.5.0"),
+    .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay.git", revision: "7bd4ee4"),
```
mbrandonw, line 22: will re-pin when this https://github.com/pointfreeco/xctest-dynamic-overlay/pull/42 is merged and released.

.github/workflows/ci.yml
```diff
@@ -19,7 +19,6 @@ jobs:
     strategy:
       matrix:
         xcode:
-          - 12.5.1
           - '13.0'
```
stephencelis, line 22: Should we update this to `macos-12` and Xcode 14.2?

Package.swift
```diff
@@ -1,27 +1,38 @@
...
     .library(
       name: "CasePaths",
       targets: ["CasePaths"]
     )
   ],
   dependencies: [
-    .package(name: "Benchmark", url: "https://github.com/google/swift-benchmark", from: "0.1.0")
+    .package(name: "Benchmark", url: "https://github.com/google/swift-benchmark", from: "0.1.0"),
+    //.package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay.git", from: "0.5.0"),
+    .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay.git", branch: "spi-current-test-case"),
```
stephencelis, line 22: We typically omit `.git` in our other packages (_e.g._ in `swift-composable-architecture`), so I think we should here too to avoid weird Package.resolved churn/conflicts.

```suggestion
    //.package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "0.5.0"),
    .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", branch: "spi-current-test-case"),
```

Tests/CasePathsTests/XCTUnwrapTests.swift
```diff
@@ -0,0 +1,30 @@
+import CasePaths
+import XCTest
+
+final class XCTUnwrapTests: XCTestCase {
```
stephencelis, line 4: Should this test case be wrapped in `#if DEBUG` since it relies on overlay?

mbrandonw: yep good catch!

stephencelis (approved):

.github/workflows/ci.yml
```diff
@@ -14,31 +14,12 @@ concurrency:
 
 jobs:
   macos-11:
```
stephencelis, line 16:
```suggestion
  macos-12:
```

Sources/CasePaths/XCTUnwrap.swift
```diff
@@ -0,0 +1,38 @@
...
+///   - expression: An enum value.
+///   - extract: A closure that attempts to extract a particular case from the enum. You can supply
+///     a case path literal here, like '/Enum.case'.
+///   - message: An optional description of a failure.
+/// - Returns: The unwrapped associated value from the matched case of the enum.
+public func XCTUnwrap<Root, Case>(
+  _ expression: @autoclosure () throws -> Root,
+  case extract: (Root) -> Case?,
+  _ message: @autoclosure () -> String = "",
+  file: StaticString = #filePath,
```
stephencelis, line 16:
```suggestion
  file: StaticString = #file,
```

PR #101, opened 2023-02-23
Introducing XCTModify

mbrandonw:
This PR introduces a more streamlined way to perform an in-place mutation of a case in an enum when in a testing context:

```swift
try XCTModify(&result, case: /Result.success) {
  $0 += 1
}
```

You should think of `XCTModify` as a companion to `XCTUnwrap` as it allows you to safely unwrap a case from an enum, and then further apply a mutation. This helper will be very useful for the navigation tools being built for TCA.


stephencelis (approved):
Looks good!

PR #109, opened 2023-04-03
Add more failure message context to XCT helpers

stephencelis:
Cleans up the APIs and messaging for `XCTModify`/`XCTUnwrap`.

### DIFF
--- a/Sources/CasePaths/EnumReflection.swift
+++ b/Sources/CasePaths/EnumReflection.swift
@@ -352,6 +352,10 @@ extension EnumMetadata {
 
     return type
   }
+
+  @_spi(Reflection) public func caseName(forTag tag: UInt32) -> String? {
+    self.typeDescriptor.fieldDescriptor?.field(atIndex: tag).name
+  }
 }
 
 @_silgen_name("swift_getTypeByMangledNameInContext")
@@ -540,6 +544,14 @@ private struct FieldRecord {
       .advanced(by: 4)
       .loadRelativePointer()
       .map { MangledTypeName(ptr: $0.assumingMemoryBound(to: UInt8.self)) }
+  }
+
+  var name: String? {
+    self.ptr
+      .advanced(by: 4)
+      .advanced(by: 4)
+      .loadRelativePointer()
+      .map { String(cString: $0.assumingMemoryBound(to: CChar.self)) }
   }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/CasePaths/EnumReflection.swift
```diff
@@ -352,6 +352,10 @@ extension EnumMetadata {
 
     return type
   }
+
+  @_spi(Reflection) public func caseName(forTag tag: UInt32) -> String? {
```
stephencelis, line 356: I added this originally to use, but decided it wasn't needed...still useful SPI, and maybe we'll tweak the messaging to use it directly instead...

Sources/CasePaths/XCTUnwrap.swift
```diff
@@ -4,28 +4,33 @@ import Foundation
...
 ///   - extract: A closure that attempts to extract a particular case from the enum. You can supply
 ///     a case path literal here, like '/Enum.case'.
 ///   - message: An optional description of a failure.
 /// - Returns: The unwrapped associated value from the matched case of the enum.
-public func XCTUnwrap<Root, Case>(
-  _ expression: @autoclosure () throws -> Root,
-  case extract: (Root) -> Case?,
+public func XCTUnwrap<Enum, Case>(
+  _ enum: @autoclosure () throws -> Enum,
+  case extract: (Enum) -> Case?,
```
stephencelis, line 14: Some renaming.

Sources/CasePaths/XCTUnwrap.swift
```diff
@@ -4,28 +4,33 @@ import Foundation
...
     XCTFail(
       """
-      XCTUnwrap failed: expected non-nil value of type "\(typeName(Case.self))"\
-      \(message.isEmpty ? "" : " - " + message)
+      XCTUnwrap failed: expected to extract value of type "\(typeName(Case.self))" from \
+      "\(typeName(Enum.self))"\
+      \(message.isEmpty ? "" : " - " + message) …
+
+        Actual:
+          \(String(describing: `enum`))
```
stephencelis, line 33: I found `String(describing:)` to be good enough for most examples. Would be nice to use custom dump, but prob not a dependency we wanna bring in just for these helpers.

mbrandonw (approved):
nice!

PR #111, opened 2023-05-17
Add CasePathed protocol

finestructure:
This is a small PR accompanying discussion #93

I found the playground where I'd played around with this while cleaning up 😅 and wanted to see if there's interest to add it!

### DIFF
--- a/Tests/CasePathsTests/CasePathedTests.swift
+++ b/Tests/CasePathsTests/CasePathedTests.swift
@@ -1,0 +1,42 @@
+import CasePaths
+import XCTest
+
+final class CasePathedTests: XCTestCase {
+
+  enum Authentication: CasePathed, Equatable {
+    case authenticated(accessToken: String)
+    case unauthenticated
+  }
+
+  func testGetSubscript() {
+    do {
+      let auth = Authentication.authenticated(accessToken: "cafebeef")
+      XCTAssertNil(auth[casePath: /Authentication.unauthenticated])
+      XCTAssertEqual(auth[casePath: /Authentication.authenticated], "cafebeef")
+    }
+    do {
+      let auth = Authentication.unauthenticated
+      XCTAssertNotNil(auth[casePath: /Authentication.unauthenticated])  // can't test == ()
+      XCTAssertNil(auth[casePath: /Authentication.authenticated])
+    }
+  }
+
+  func testSetSubscript() {
+    do {
+      var auth = Authentication.authenticated(accessToken: "")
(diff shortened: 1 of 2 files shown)


Tests/CasePathsTests/CasePathedTests.swift
```diff
@@ -0,0 +1,42 @@
...
+  func testSetSubscript() {
+    do {
+      var auth = Authentication.authenticated(accessToken: "")
+      auth[casePath: /Authentication.authenticated] = "cafebeef"
+      XCTAssertEqual(auth, .authenticated(accessToken: "cafebeef"))
+    }
+    do {
+      var auth = Authentication.unauthenticated
+      auth[casePath: /Authentication.authenticated] = "cafebeef"
+      XCTAssertEqual(auth, .authenticated(accessToken: "cafebeef"))
```
finestructure, line 33: Is this weird that it changes the whole enum instead of just the associated value (which it couldn't in this case)?

I suppose it's ok, because in normal use you'd only really want to use the subscript to change the associated value since you can simply assign `auth = .authenticated("cafebeef")` otherwise.

It might be a bit weird if you have

```swift
  enum A: CasePathed, Equatable {
    case b(Int)
    case c(Int)
  }
```

and then do

```swift
var a = A.b(1)
a[casePath: /A.c] = 2
XCTAssertEqual(a, .c(2))
```

🤔

iampatbrown: Hey @finestructure, thanks for submitting this PR. 

This is definitely something that's desirable and others have attempted to address this in the past (see #62 and #110). The main challenge is being able to have an optional getter and non-optional setter. eg.

```swift
let a = A.b(1)
a[casePath: /A.c] = 2 // ✅
a[casePath: /A.c] = nil // 🛑 forbidden by the compiler
```

@stephencelis provides some more context [here](https://github.com/pointfreeco/swift-case-paths/pull/62#issuecomment-998024589).

The last time we experimented with this we got pretty close but it involved a bit of trickery to work around Swift's current limitations. Feel free to check it out here https://github.com/pointfreeco/swift-case-paths/compare/main...iampatbrown:swift-case-paths:case-accessible

finestructure: Oh, I see! I had no idea this was such a well worked field - but it's kinda obvious in hindsight :)

Looks like you're all way ahead of me on this one, as is to be expected. I'll go ahead and close this then, thanks for the pointers!

PR #113, opened 2023-05-24
Breaking change to CasePath.modify

stephencelis:
We'd like to make a breaking change to `CasePath.modify`:

```diff
-func modify<R>(inout Enum, (inout Case) throws -> R) throws -> R
+func modify(inout Enum, (inout Case) throws -> Void) rethrows
```

1. It no longer throws (unless the transform closure throws)
2. It no longer returns a result through the continuation.

For 1., we're not sure error handling in the signature makes sense. Modifying the wrong case seems more like a programmer error than anything else. So instead, we'll runtime warn.

For 2., losing `throws` means the result would need to be optionalized. While this is possible, we also don't think values are typically assigned through the result of this closure, so let's remove support for this completely.

For most folks we believe this change will only introduce a build warning:

> **Warning**: No calls to throwing functions occur within 'try' expression

So hopefully not too disruptive!

### DIFF
--- a/Sources/CasePaths/CasePath.swift
+++ b/Sources/CasePaths/CasePath.swift
@@ -1,4 +1,5 @@
 import Foundation
+import XCTestDynamicOverlay
 
 /// A path that supports embedding a value in a root and attempting to extract a root's embedded
 /// value.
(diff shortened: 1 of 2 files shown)


lukaskubanek:
I noticed a minor oversight in your comment above. Based on the description of the change and the associated commit, the new signature of the method should return `Void` instead of `R`.

Also, the `ExtractionFailed` type can probably be removed from the codebase.

stephencelis:
Going to close this for now, as we have plans to revisit the functionality in a different way soon.

PR #116, opened 2023-09-01
Adding Windows CI

brianmichel:
We want to setup Windows CI for all of the Pointfree repos that are used in TCA to ensure that we have cross platform compatibility going forward (or at least to ensure that nothing breaks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,5 @@ jobs:
           tag: 5.8.1-RELEASE
 
       - uses: actions/checkout@v3
-      - name: Build
-        run: swift build -c ${{ matrix.config }}
-      - name: Test
-        run: swift test --skip-build
+      - name: Build & Test
+        run: swift test -c ${{ matrix.config }}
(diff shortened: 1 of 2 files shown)


brianmichel:
All of these tests currently pass on Windows!

brianmichel:
Ran these tests locally with the 5.8.1 toolchain which is the most non-development current toolchain and everything seems to pass there so I bumped the config to that.

mbrandonw:
Hi @brianmichel, thanks for this! If you can't figure out why 5.8.1 is still failing on CI, feel free to bump to 5.9 if that gets things passing.

brianmichel:
@mbrandonw yeah it's on my list this week, just getting sidetracked by some C++ Interop stuff that's blocking our project which took precedence. 🤞 it's an easy fix to get this to pass!

brianmichel:
Hey @mbrandonw looks like there was some bug preventing running the tests in a release config, so I've broken out the build phases and the test phases to try to get a good balance of coverage on the profiles.

brianmichel:
@mbrandonw looks like this should be good now, had to make a couple adjustments but left comments where I did

mbrandonw (approved):
Hi @brianmichel looks great, thanks for figuring out the edge cases! Once Swift 5.9 is available on Windows we can revisit bringing back release testing too.

PR #117, opened 2023-09-06
Update CasePathableMacro to SwiftSyntax's latest API

Ryu0118:


mbrandonw:
Hi @Ryu0118, thanks for taking the time to do this, but this is actually an old experimental branch that we aren't using anymore. I'm going to close this for now.

PR #118, opened 2023-09-28
📝 Fix doc note of non-void `/` operators

mrackwitz:
Pretty sure these doc notes were just accidentally copy-pasted from the void variant of the operator, because these two deal with actual values. I replaced it with a note from elsewhere documenting that these should be used indeed only with case initializers.


stephencelis (approved):
Thanks!

PR #122, opened 2023-11-10
Add `@CasePathable` to any nested enums of a `@CasePathable` enum.

oliverfoggin:
This PR adds the `@CasePathable` attribute to any nested enums within an enum.

This is important for TCA as many people use "Domain" actions like...

```
enum Action {
  case view(ViewAction)
  case delegate(DelegateAction)

  enum ViewAction {
    case buttonTapped
  }

  enum DelegateAction {
    case doSomething
  }
}
```

Without this, the `\.view` and `\.delegate` are case pathable but then the only option after that is `.never`.

Adding this will allow for `\.view.buttonTapped` and `\.delegate.doSomething`.

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -44,6 +44,22 @@ extension CasePathableMacro: ExtensionMacro {
   }
 }
 
+extension CasePathableMacro: MemberAttributeMacro {
+	public static func expansion(
+		of node: SwiftSyntax.AttributeSyntax,
+		attachedTo declaration: some SwiftSyntax.DeclGroupSyntax,
+		providingAttributesFor member: some SwiftSyntax.DeclSyntaxProtocol,
+		in context: some SwiftSyntaxMacros.MacroExpansionContext
+	) throws -> [SwiftSyntax.AttributeSyntax] {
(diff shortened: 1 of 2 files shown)


stephencelis:
@oliverfoggin Thanks for the PR! We would love this functionality, but when we tried it ourselves, despite the test macro expanding, we encountered a compiler error when actually wiring a nested enum up, and reported the error here:

https://github.com/apple/swift/issues/69759

I think this PR is missing the code that specifies the member attribute macro in the `CasePaths` module, and I believe you may encounter the same error we did after doing so and attempting to use the macro 😞

oliverfoggin:
@stephencelis ah, that's a bummer. I figured you would have added it when I "thought" how easy it was. But yeah, makes sense that you didn't.

JimRoepcke:
@oliverfoggin @stephencelis rather than applying the macro to the nested enums for recursive expansion, is it possible for the top level macro expansion to recursively generate the code for the nested enums?

stephencelis:
@JimRoepcke Sadly I don't think that works since it needs to do member macro work on that type and the only way to do that is by applying another macro to it.

It is possible to recursively apply a brand new macro, so we could alias `@CasePathable` to `@CasePathable2` and so on, but that seems like a hack. Hopefully Apple will fix the limitation, so feel free to watch the above issue!

PR #128, opened 2023-11-14
Modify isNeededAccessLevelModifier result conditions

nnsnodnb:
When using the [`package`](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md) access modifier, a compile error occurred due to differences in property access levels.

Before

`package` access modifier is now given a modifier as well as public access modifier.

After


stephencelis (approved):
Thanks! Hadn't gotten around to taking that feature for a spin.

PR #130, opened 2023-11-15
Fix compile error caused by wildcard argument usage

hj56775:
I ran into a compile error `"Tuple element cannot have two labels"` with some wildcard arguments in enum cases during our TCA 1.4.0 migration. 
I've sorted it out by setting firstName to nil in associatedValue.parameter for those wildcards.
If you've got a neater way to handle this, I'm all ears for suggestions!

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -257,6 +257,9 @@ extension EnumCaseElementListSyntax.Element {
         for index in associatedValue.parameters.indices {
           associatedValue.parameters[index].type.trailingTrivia = ""
           associatedValue.parameters[index].defaultValue = nil
+          if associatedValue.parameters[index].firstName?.tokenKind == .wildcard {
+            associatedValue.parameters[index].firstName = nil
+          }
         }
         return "(\(associatedValue.parameters.trimmed))"
       }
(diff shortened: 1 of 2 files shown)


Sources/CasePathsMacros/CasePathableMacro.swift
```diff
@@ -257,6 +257,9 @@ extension EnumCaseElementListSyntax.Element {
         for index in associatedValue.parameters.indices {
           associatedValue.parameters[index].type.trailingTrivia = ""
           associatedValue.parameters[index].defaultValue = nil
+          if associatedValue.parameters[index].firstName?.tokenKind == .wildcard {
+            associatedValue.parameters[index].firstName = nil
```
stephencelis, line 261: I tried to push to your branch but looks like your fork doesn't have those permissions.

I think wildcard parameters should not propagate the second name into the final tuple type, so wanna change this and re-record the snapshot?

```suggestion
            associatedValue.parameters[index].color = nil
            associatedValue.parameters[index].firstName = nil
            associatedValue.parameters[index].secondName = nil
```

stephencelis (commented):
Thanks for the PR! Would love to address this edge case. Just one small change.

Tests/CasePathsMacrosTests/CasePathableMacroTests.swift
```diff
@@ -299,6 +299,40 @@ final class CasePathableMacroTests: XCTestCase {
...
+      }
+      """
+    } expansion: {
+      """
+      enum Foo {
+        case bar(_ int: Int, _ bool: Bool)
+
+        struct AllCasePaths {
+          var bar: CasePaths.AnyCasePath<Foo, (int: Int, bool: Bool)> {
+            CasePaths.AnyCasePath<Foo, (int: Int, bool: Bool)>(
```
stephencelis, line 317:
```suggestion
          var bar: CasePaths.AnyCasePath<Foo, (Int, Bool)> {
            CasePaths.AnyCasePath<Foo, (Int, Bool)>(
```

PR #132, opened 2023-11-27
Export CasePathsMacros

ImSingee:


mbrandonw:
Hi @ImSingee, can you explain a bit why you need access to the `CasePathMacros` module? Are you building a tool that uses the various macro conformances somehow? And if so, can you share more details?

We would be happy to publicize this module if it's useful. We just aren't aware of any actual uses for it.

ImSingee:
@mbrandonw Ah I just think it's necessary.

I can't use `@CasePathable` macros... It always show an error `External macro implementation type 'CasePathsMacros.CasePathableMacro' could not be found for macro 'CasePathable()'`

mbrandonw:
That should work fine. There's no need to publicize the macros module. Can you share a minimal reproducing project that demonstrates the behavior you are seeing?

ImSingee:
@mbrandonw So strange😂 I just remove my xcuserdata and then it works.

Sorry for wasting your time...

PR #133, opened 2023-12-05
Support enums with cases wrapped in #if

AndrewVebster:
## Summary

This PR updates the macro code to allow it to generate case paths for cases which are conditionally compiled. For example:
```
@CasePathable enum Foo {
  case bar
  #if DEBUG
  case debugCase
  #endif
}
```

_Currently released implementation would generate the case path only for `bar`._

## Why

I'm working on a project where we use TCA. Because we share the code between iOS and macOS we have a few places where one platform has some additional actions (features) and they are wrapped with `#if os`.

What do you think? Would it be useful? Appreciate any feedback 🙂

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -87,13 +87,53 @@ extension CasePathableMacro: MemberMacro {
       seenCaseNames.insert(name)
     }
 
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Awesome addition! Thanks for taking the time to dive into the problem and write some great tests :smile:

AndrewVebster:
Happy to contribute 😉 🙌

PR #134, opened 2023-12-13
Support matching on partial key paths using `is`

lukeredpath:
Updated the signature to be a `PartialCaseKeyPath` which is all that is required for this function to work.

Makes it possible to write algorithms that operate on a collection of case key paths for a given type, for example:

```swift
extension CasePathable {
  func isOneOf<CaseKeyPaths: Sequence>(_ keyPaths: CaseKeyPaths) -> Bool where CaseKeyPaths.Element == PartialCaseKeyPath<Self> {
    keyPaths.contains(where: { self.is($0) })
  }
}
```


mbrandonw (approved):
Awesome, thanks!

PR #140, opened 2024-01-04
Simplify optional flattening

stephencelis:
Turns out we only need the `Case` overload.

Also documenting a few paths while I go.


Sources/CasePaths/Optional+CasePathable.swift
```diff
@@ -42,6 +29,10 @@ extension Optional: CasePathable {
 }
 
 extension Case {
+  /// A case path to the presence of a nested value.
+  ///
+  /// This subscript can chain into an optional's wrapped value without explicitly specifying each
+  /// `none` component.
```
stephencelis, line 35:
```suggestion
  /// `some` component.
```

mbrandonw (approved):

PR #150, opened 2024-03-04
Allow writability when chaining into a dynamic case

stephencelis:
While we should prohibit directly assigning a value through a case:

```swift
state.loggedIn = User(name: "Blob")  // 🛑
```

We can allow optional-chaining into a case's sub-value via an overload and some disfavoring:

```swift
state.loggedIn?.name = "Blob"  // ✅
```

### DIFF
--- a/Tests/CasePathsTests/CasePathsTests.swift
+++ b/Tests/CasePathsTests/CasePathsTests.swift
@@ -70,7 +70,8 @@ final class CasePathsTests: XCTestCase {
       var foo: Foo = .bar(.int(1))
 
       XCTAssertEqual(foo.bar, .int(1))
-      XCTAssertEqual(foo.bar?.int, 1)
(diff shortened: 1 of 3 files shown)


Tests/CasePathsTests/CasePathsTests.swift
```diff
@@ -70,7 +70,8 @@ final class CasePathsTests: XCTestCase {
       var foo: Foo = .bar(.int(1))
 
       XCTAssertEqual(foo.bar, .int(1))
-      XCTAssertEqual(foo.bar?.int, 1)
+      // NB: Due to a Swift bug, this is only possible to do outside the library:
+      // XCTAssertEqual(foo.bar?.int, 1)
```
stephencelis, line 74: This seems to trigger a Swift bug in the package in which the unavailable setter is accessed. Outside the package things appear to resolve just fine, though!

PR #154, opened 2024-04-05
leadingTrivia indentation trimming calculation should not discard zero values

djangovanderheijden:
With v1.3.1 documentation is now also included for the generated `AllCasePaths` struct.
In order to achieve this the leadingTrivia for each case is appended to the generated properties. To ensure indentation matches, the code discards leading whitespaces from the trivia.
When determining how many characters to trim, the code loops over each line in the leadingTrivia, discarding empty lines and counting the amount of prefixed whitespaces. The smallest value counted is then taken and used as the amount of whitespace to trim from each line.

There is an issue in this last bit of logic where lines without prefixed whitespaces are always ignored. This will cause these lines to be trimmed erroneously. Take the following example:
```swift
@CasePathable
enum MyEnum {
    // TODO: Implement this case
//    case one
    case two
}
```

Because the line `//    case one` is ignored when calculating the minimum amount of whitespace, it is trimmed in the generated code:

```swift
public struct AllCasePaths {
    // TODO: Implement this case
      case one
    public var two: CasePaths.AnyCasePath<MyEnum, Void> {
        ...
    }

    …
}
```

This obviously causes a build failure.

The change removes the bit of logic where lines with zero whitespacing are ignored when calculating the minimum amount of whitespace. As far as I can tell this doesn’t have a negative impact. Tests still pass including`testDocumentation`.

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -143,7 +143,7 @@ extension CasePathableMacro: MemberMacro {
       let indent =
         leadingTriviaLines
         .compactMap { $0.isEmpty ? nil : $0.prefix(while: \.isWhitespace).count }
-        .min(by: { (lhs: Int, rhs: Int) -> Bool in lhs == 0 ? lhs > rhs : lhs < rhs })
+        .min(by: { (lhs: Int, rhs: Int) -> Bool in lhs < rhs })
         ?? 0
       let leadingTrivia =
         leadingTriviaLines
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -149,6 +149,7 @@ extension CasePathableMacro: MemberMacro {
         leadingTriviaLines
         .map { String($0.dropFirst(indent)) }
         .joined(separator: "\n")
+        .trimmingSuffix(while: { $0.isWhitespace && !$0.isNewline })
       return """
         \(raw: leadingTrivia)public var \(caseName): \
         \(raw: qualifiedCasePathTypeName)<\(enumName), \(raw: associatedValueName)> {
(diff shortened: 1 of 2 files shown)


djangovanderheijden:
The underlying problem here is that Swift considers the first commented out case to be trivia relating to the second case even though it isn't. I don't think that's easy to fix though :)

stephencelis:
@djangovanderheijden Thanks! Any chance you can add a macro test case for this? Would be good to capture the fix in test coverage.

djangovanderheijden:
Hey @stephencelis, no problem! I added a small test that fails in the old situation and passes with the changes made.

I also noticed that the indentation of the line on which the property itself is declared (e.g. `public var two: CasePaths.AnyCasePath<MyEnum, Void> {`) was slightly different from the indentation found in other tests with the change. To combat this, I added a small tweak that removes trailing whitespaces excluding newlines from the very end of the leadingTrivia. All tests still pass and the change seems harmless enough so I figured I'd include it.

Sources/CasePathsMacros/CasePathableMacro.swift
```diff
@@ -143,12 +143,13 @@ extension CasePathableMacro: MemberMacro {
...
         leadingTriviaLines
         .compactMap { $0.isEmpty ? nil : $0.prefix(while: \.isWhitespace).count }
-        .min(by: { (lhs: Int, rhs: Int) -> Bool in lhs == 0 ? lhs > rhs : lhs < rhs })
+        .min(by: { (lhs: Int, rhs: Int) -> Bool in lhs < rhs })
         ?? 0
       let leadingTrivia =
         leadingTriviaLines
         .map { String($0.dropFirst(indent)) }
         .joined(separator: "\n")
+        .trimmingSuffix(while: { $0.isWhitespace && !$0.isNewline })
```
djangovanderheijden, line 152: This ensures trailing whitespaces are trimmed from the trivia. It ensures no whitespaces are added after the final newline, potentially messing up the indentation of the property declaration.

Sources/CasePathsMacros/CasePathableMacro.swift
```diff
@@ -422,3 +423,16 @@ final class SelfRewriter: SyntaxRewriter {
     return super.visit(node.with(\.name, self.selfEquivalent))
   }
 }
+
+extension StringProtocol {
```
djangovanderheijden, line 427: I wish Swift would have a built-in equivalent to `trimmingPrefix(while)`, but this will do

stephencelis (approved):
Thanks!

PR #155, opened 2024-04-05
CasePathable.modify: Change XCTFail to runtimeWarn as per documentation

djangovanderheijden:
As per [the documentation](https://swiftpackageindex.com/pointfreeco/swift-case-paths/main/documentation/casepaths/casepathable/modify(_:yield:file:line:)):
> If the enum’s case does not match the given case key path, the mutation will not be applied, and a runtime warning will be logged.

In reality, calling `modify` with a non-matching case will always call `XCTFail`. This PR adds ComposableArchitecture's `RuntimeWarnings.swift` file and uses `runtimeWarn` instead of `XCTFail`.

Alternatively, we could consider changing the documentation to clarify that the use of a non-matching case causes a runtime crash. I'm open to either solution, but a warning log seems a bit more graceful.

### DIFF
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -479,13 +479,11 @@ extension CasePathable {
   ) {
     let `case` = Case(keyPath)
     guard var value = `case`.extract(from: self) else {
-      XCTFail(
+      runtimeWarn(
         """
         Can't modify '\(String(describing: self))' via 'CaseKeyPath<\(Self.self), \(Value.self)>' \
         (aka '\(String(reflecting: keyPath))')
-        """,
-        file: file,
-        line: line
+        """
       )
       return
     }
--- a/Sources/CasePaths/Internal/RuntimeWarnings.swift
+++ b/Sources/CasePaths/Internal/RuntimeWarnings.swift
@@ -1,0 +1,85 @@
+import Foundation
+
+extension Notification.Name {
+  #if swift(>=5.8)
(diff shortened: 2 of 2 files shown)


mbrandonw:
Hi @djangovanderheijden, our `XCTFail` from xctest-dynamic-overlay does a [runtime warning](https://github.com/pointfreeco/xctest-dynamic-overlay/blob/5ede785843718dc6619b57b22a735bed541de7c9/Sources/XCTestDynamicOverlay/XCTFail.swift#L42-L44) under the hood when not in tests.

So this does currently behave as described in the docs. Are you seeing otherwise? Also, this definitely should not _crash_ the app. Are you seeing that?

djangovanderheijden:
Hey @mbrandonw, interesting. I did see a crash in my application (with the debugger attached, but not while testing). I'll take another look and get back to you

djangovanderheijden:
Alright, the issue seems unrelated to CasePaths (sorry about that, I should've read the error on my colleague's screen better) but I am definitely seeing crashes:
Screenshot 2024-04-05 at 20 08 24

This happens in a clean, newly created project that's ran for the first time. Here's a small repro:
```swift
@CasePathable
enum Foo {
    case bar(String)
    case baz
}

struct ContentView: View {
    var body: some View {
        Text("Foo")
            .onAppear {
                var foo = Foo.baz
                foo.modify(\.bar) {
                    $0 = "bar"
                }
            }
    }
}
```

djangovanderheijden:
Nice! I'll close this.

PR #157, opened 2024-04-26
Fix Optional dynamic member lookup

stephencelis:
Swift optional promotion is sometimes succeeding to return `.some(.none)` where it should be returning `.none`.

### DIFF
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -440,12 +440,7 @@ extension CasePathable {
   /// userActions.filter { $0.is(\.home) }      // [UserAction.home(.onAppear)]
   /// userActions.filter { $0.is(\.settings) }  // [UserAction.settings(.subscribeButtonTapped)]
   /// ```
-  @_disfavoredOverload
   public func `is`(_ keyPath: PartialCaseKeyPath<Self>) -> Bool {
-    self[case: keyPath] != nil
-  }
-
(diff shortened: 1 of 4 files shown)


Sources/CasePaths/CasePathable.swift
```diff
@@ -440,15 +440,10 @@ extension CasePathable {
...
   /// userActions.filter { $0.is(\.settings) }  // [UserAction.settings(.subscribeButtonTapped)]
   /// ```
-  @_disfavoredOverload
   public func `is`(_ keyPath: PartialCaseKeyPath<Self>) -> Bool {
     self[case: keyPath] != nil
   }
 
-  public func `is`<Wrapped>(_ keyPath: CaseKeyPath<Self, Wrapped?>) -> Bool {
-    self[case: keyPath] != nil
-  }
```
stephencelis, line 450: Unrelatedly, we found that this overload is no longer necessary with some upstream optional promotion fixes.

stephencelis:
Thanks @dannyhertz for the report!

mbrandonw (approved):

Tests/CasePathsTests/CasePathsTests.swift
```diff
@@ -182,6 +182,14 @@ final class CasePathsTests: XCTestCase {
       XCTAssertNotNil(foo(String?.some("Blob") as Any))
       XCTAssertNotNil(foo("Blob"))
     }
+
+    func testIs_Optional() {
```
dannyhertz, line 186: Testable AND tested! We love to see it, folks!

dannyhertz (approved):

PR #159, opened 2024-05-31
Add case path iteration

stephencelis:
This wasn't so bad to add on top of #158.

I made it so that `AllCasePaths: Sequence` so that you could iterate over all of the case paths.

The biggest issue is this conformance confuses autocompletion of case key paths with a few properties that come with `Sequence`:

image

They show up at the end, so it's not so bad. I avoided a `RandomAccessCollection` conformance because it added even more pollution here.

Alternately:

1. We could spell out a whole separate object that acts as the collection, though not sure what that would look like.
2. If/when we add some kind of conformance to `AllCasePaths`, like `CasePathCollection`, then we could squirrel away the collection stuff to a `RangeReplaceableCollection.init(some CasePathCollection)` helper instead.


mbrandonw (approved):

stephencelis:
(Restored the branch for https://github.com/apple/swift/issues/74107)

PR #160, opened 2024-06-06
Handle comments on cases.

mbrandonw:


Tests/CasePathsMacrosTests/CasePathableMacroTests.swift
```diff
@@ -968,4 +968,102 @@ final class CasePathableMacroTests: XCTestCase {
...
+        public struct AllCasePaths: Sequence {
+          public subscript(root: Foo) -> PartialCaseKeyPath<Foo> {
+            switch root {
+            case .bar:
+              return \.bar
+            case .baz:
+              return \.baz
+            case .fizz:
+              return \.fizz
+            case .fizzier/*Comment in case*/:
```
mbrandonw, line 1001: we could even trim this comment, but it's also not hurting anything and just an edge case

Tests/CasePathsTests/MacroTests.swift
```diff
@@ -0,0 +1,10 @@
+import CasePaths
+
+@CasePathable
+private enum Comments {
+  // Comment above case
+  case bar
+  /*Comment before case*/ case baz(Int)
+  case fizz(buzz: String)  // Comment on case
+  case fizzier/*Comment in case*/(Int, buzzier: String)
+}
```
mbrandonw, line 10: added a "static" macro test. we could probably have more of these.

stephencelis (approved):

PR #163, opened 2024-06-07
Handle comments on cases without associated values

twocentstudios:
- https://pointfreecommunity.slack.com/archives/C04KQQ7NXHV/p1717736207509849?thread_ts=1717708594.669469&cid=C04KQQ7NXHV
- https://github.com/pointfreeco/swift-case-paths/pull/160/

Fixes compilation error for enum cases without associated values that have trailing comments e.g.:

```swift
// Using swift-case-paths 1.4.1
@CasePathable enum Foo {
  case fizz(buzz: String) // Comment on case ✅
  case fizziest // Comment without associated value ❌ 
}
```

Previously, for trailing comments on enum cases without associated values, the macro was generating invalid syntax in two places:

```swift
// public struct AllCasePaths: Sequence {
case .fizziest // Comment without associated value: return \.fizziest ❌

// public func makeIterator() -> IndexingIterator<[PartialCaseKeyPath<Foo>]> {
allCasePaths.append(\.fizziest // Comment without associated value) ❌
```

I used the same fix from #160 – adding `.text`. This has the side-effect of removing C-style comments outside the root enum declaration as mentioned in [this comment](https://github.com/pointfreeco/swift-case-paths/pull/160#discussion_r1630303847), which I believe is not a problem.

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -84,7 +84,7 @@ extension CasePathableMacro: MemberMacro {
     let rewriter = SelfRewriter(selfEquivalent: enumName)
     let memberBlock = rewriter.rewrite(enumDecl.memberBlock).cast(MemberBlockSyntax.self)
     let rootSwitchCases = generateCases(from: memberBlock.members, enumName: enumName) {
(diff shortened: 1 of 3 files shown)


twocentstudios:
(I'm not particularly versed in writing macros, so feel free to close this PR if you have a better solution 🙆‍♂️)

PR #164, opened 2024-06-07
Fix build warning when `@CasePathable` is applied to enums with no cases

jpsim:
Fixes warnings due to unreachable code introduced in [1.4.0](https://github.com/pointfreeco/swift-case-paths/releases/tag/1.4.0):

For example:

```swift
@CasePathable enum Foo {}
```

Produces this expansion (compiler warnings added as code comments by me)

```swift
enum Foo: CasePathable {
  public struct AllCasePaths: Sequence {
    public subscript(root: Foo) -> PartialCaseKeyPath<Foo> {
      return \.never // ⚠️ warning: will never be executed
      // 'root' is of type 'Foo' which cannot be constructed because it
      // is an enum with no cases
    }

    public func makeIterator() -> IndexingIterator<[PartialCaseKeyPath<Foo>]> {
      let allCasePaths: [PartialCaseKeyPath<Foo>] = []
      return allCasePaths.makeIterator()
    }
  }
  public static var allCasePaths: AllCasePaths { AllCasePaths() }
}
```

We can fix this warning by simply not producing the subscript definition when the enum has no cases.

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -91,14 +91,14 @@ extension CasePathableMacro: MemberMacro {
       "allCasePaths.append(\\.\(raw: $0.name.text))"
     }
 
+    let subscriptReturn = allCases.isEmpty ? #"\.never"# : #"return \.never"#
+
     return [
       """
       public struct AllCasePaths: Sequence {
-      \(allCases.isEmpty ? "" : """
-        public subscript(root: \(enumName)) -> PartialCaseKeyPath<\(enumName)> {
-        \(raw: rootSubscriptCases.map { "\($0.description)\n" }.joined())return \\.never
-        }
-        """)
+      public subscript(root: \(enumName)) -> PartialCaseKeyPath<\(enumName)> {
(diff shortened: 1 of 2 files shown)


stephencelis:
Hey @jpsim! For some reason these diagnostics are really thorny, and probably warrant some bug reports to Swift. For example, this compiles without warning:

```swift
func never<A>(never: Never) -> A {}

enum Foo {}
func never<A>(never: Foo) -> A {}
```

So I think technically this should compile, but it doesn't:

```swift
enum Foo {
  static subscript(root: Foo) -> PartialKeyPath<Foo> {
}
```

I'll file a bug soon.

Ideally we'd be able to have the implementation of this subscript regardless, since we eventually want to abstract it on the `CasePathable` protocol. We're just delaying that work, probably to 2.0, since it feels like a breaking change to folks with manual implementations.

I have found that omitting the `return` statement and still returning `\.never` silences the warning, though:

```diff
 public subscript(root: Foo) -> PartialCaseKeyPath<Foo> {
-  return \.never
+  \.never
```

Can you try making that change and confirm?

jpsim:
Thanks for the details! I'll give that a shot.

stephencelis:
https://github.com/apple/swift/issues/74210

stephencelis (approved):

PR #168, opened 2024-06-13
Swift 6 Language Mode

stephencelis:
This PR adds support for Swift 6 and improves the sendability of case paths.


mbrandonw (approved):

stephencelis:
@brianmichel @compnerd We've unfortunately had to disable the Windows CI in this PR due to compiler assertion failures and other inscrutable build failures while trying to debug them. We'd be happy to take PRs to address the issues if someone with more Windows experience has time to take a look, but we've already spent a significant amount of time trying to get to the bottom of things. I've mentioned it before, but as library maintainers it might make sense to have a readymade option for Windows toolchains that have assertions disabled so that CI can at least be consistent with the other toolchains we use, and so that we don't get stuck with a red build failure due to a Swift debug assertion.

brianmichel:
> @brianmichel @compnerd We've unfortunately had to disable the Windows CI in this PR due to compiler assertion failures and other inscrutable build failures while trying to debug them. We'd be happy to take PRs to address the issues if someone with more Windows experience has time to take a look, but we've already spent a significant amount of time trying to get to the bottom of things. I've mentioned it before, but as library maintainers it might make sense to have a readymade option for Windows toolchains that have assertions disabled so that CI can at least be consistent with the other toolchains we use, and so that we don't get stuck with a red build failure due to a Swift debug assertion.

No worries @stephencelis can you maybe cut an issue on this repo and assign it to me, I don't have tons of free time, but am happy to play around and see if I can reduce the issue you're hitting into something workable.

Out of curiosity do you see this issue on Linux if you update the toolchain there as well?

stephencelis:
@brianmichel Will do! The linux CI is passing, but I assume because the toolchain probably doesn't have debug assertions enabled.

PR #173, opened 2024-06-21
Add `CasePathIterable` and `CasePathReflectable` protocols 

stephencelis:
While the macro recently introduced iteration (https://github.com/pointfreeco/swift-case-paths/pull/159) and "reflection" (https://github.com/pointfreeco/swift-case-paths/pull/158), there's no way to abstract over it. Adding it directly to the `CasePathable` protocol would be potentially source breaking, so instead we can introduce some new protocols.

### DIFF
--- a/Sources/CasePaths/CasePathIterable.swift
+++ b/Sources/CasePaths/CasePathIterable.swift
@@ -1,0 +1,2 @@
+public protocol CasePathIterable: CasePathable
+where AllCasePaths: Sequence, AllCasePaths.Element == PartialCaseKeyPath<Self> {}
--- a/Sources/CasePaths/CasePathIterable.swift
+++ b/Sources/CasePaths/CasePathIterable.swift
@@ -1,2 +1,17 @@
+/// A type that provides a collection of all of its case paths.
+///
+/// The ``CasePathable()`` macro automatically generates a conformance to this protocol.
+///
+/// You can iterate over ``CasePathable/allCasePaths`` to get access to each individual case path:
+///
+/// ```swift
+/// @CasePathable enum Field {
+///   case title(String)
+///   case body(String
+///   case isLive
+/// }
+///
+/// Array(Field.allCasePaths)  // [\.title, \.body, \.isLive]
+/// ```
 public protocol CasePathIterable: CasePathable
 where AllCasePaths: Sequence, AllCasePaths.Element == PartialCaseKeyPath<Self> {}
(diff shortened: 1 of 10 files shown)


Sources/CasePaths/CasePathIterable.swift
```diff
@@ -0,0 +1,2 @@
+public protocol CasePathIterable: CasePathable
```
stephencelis, line 1: - [ ] Docs

Sources/CasePaths/CasePathReflectable.swift
```diff
@@ -0,0 +1,4 @@
+public protocol CasePathReflectable<Root> {
```
stephencelis, line 1: - [ ] Docs

Tests/CasePathsTests/CaseSetTests.swift
```diff
@@ -0,0 +1,202 @@
+#if swift(>=6)
+  import CasePaths
+  import Testing
+
+  @dynamicMemberLookup
+  public struct CaseSet<Element: CasePathable> {
```
stephencelis, line 6: - [ ] Consider bringing this into the library proper

stephencelis:
@mbrandonw I think this is ready to go. The main thing is if we're happy with the naming. `CasePathIterable` seems non-controversial, but the `CasePathReflectable` naming may be a little strange given that it's defined on the `AllCasePaths` type. One might assume/consider that the root enum is the "reflectable" type, and the `AllCasePaths` is just doing the reflecting...so is there a better name? `CasePathReflecting` maybe? Though I can't think of any protocols with an "-ing" suffix off the top of my head.

Tests/CasePathsMacrosTests/CasePathableMacroTests.swift
```diff
@@ -31,7 +31,7 @@ final class CasePathableMacroTests: XCTestCase {
         case fizz(buzz: String)
         case fizzier(Int, buzzier: String)
 
-        public struct AllCasePaths: Sendable, Sequence {
+        public struct AllCasePaths: CasePathReflectable, Hashable, Sendable, Sequence {
```
stephencelis, line 34: - [ ] Remove `Hashable` for now

Tests/CasePathsMacrosTests/CasePathableMacroTests.swift
```diff
@@ -126,7 +126,7 @@ final class CasePathableMacroTests: XCTestCase {
       #"""
       enum EnumWithNoCases {
 
-          public struct AllCasePaths: Sendable, Sequence {
+          public struct AllCasePaths: CasePathReflectable, Hashable, Sendable, Sequence {
```
stephencelis, line 129: - [ ] Qualify.

mbrandonw (approved):

PR #181, opened 2024-07-15
Disable sendable dynamic key path lookup for now

stephencelis:
A regression in recent snapshots is causing compiler crashes.

https://github.com/swiftlang/swift/issues/75244

Let's disable these code paths for now while we wait for resolution.


stephencelis:
Before merging/releasing we should try to see if this produces any bad warnings downstream.

mbrandonw (approved):

PR #185, opened 2024-07-19
fix for Element compiler issue

iwt-dahlborn:
This would be the general idea with a very rough implementation. Basically just generating a `typealias` for every associated value that contains an `Element` type.

It could be refined to check if there is specifically a type named `Element` used since this will just catch anything containing element. 

I think the macro definition can also better specify what names the `typealias` can have so it doesnt need to be arbitrary?

Like I said just a very rough idea.

edit: Sorry just saw that my formatter ran a bit wild there

### DIFF
--- a/Sources/CasePathsMacros/CasePathableMacro.swift
+++ b/Sources/CasePathsMacros/CasePathableMacro.swift
@@ -86,12 +86,13 @@ extension CasePathableMacro: MemberMacro {
       seenCaseNames.insert(name)
     }
 
-    let rewriter = SelfRewriter(selfEquivalent: enumName)
-    let memberBlock = rewriter.rewrite(enumDecl.memberBlock).cast(MemberBlockSyntax.self)
+    let selfRewriter = SelfRewriter(selfEquivalent: enumName)
+    let memberBlock = selfRewriter.rewrite(enumDecl.memberBlock).cast(MemberBlockSyntax.self)
     let rootSubscriptCases = generateCases(from: memberBlock.members, enumName: enumName) {
       "if root.is(\\.\(raw: $0.name.text)) { return \\.\(raw: $0.name.text) }"
     }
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hey @iwt-dahlborn, how do you feel about tweaking this PR slightly so that rather than introducing a bunch of type aliases for each payload we just stick with the one `_$Element` type alias, and then replace occurrences of `Element` in the payloads with `_$Element`:

```diff
-public var element: CasePaths.AnyCasePath<Action, _$elementElement> {
+public var element: CasePaths.AnyCasePath<Action, (_$Element, Int)> {
```

stephencelis (commented):
Thanks for diving into this!

I think ideally we would avoid the `arbitrary` naming in the macro and preserve the simpler `_$Element` type alias if possible. Any reason this can't be done?

I was thinking we could detect an `Element` type referenced inside `AllCasePaths` and we could replace them all with `_$Element`. There's already a `SelfRewriter` that replaces instances of `Self` with the enum name, and I think we could generalize it to do this job as well:

https://github.com/pointfreeco/swift-case-paths/blob/main/Sources/CasePathsMacros/CasePathableMacro.swift#L471-L483

And then if it rewrites at least one `Element`, the macro would also insert the type alias in the parent.

Wanna give that a shot?

iwt-dahlborn:
I updated the pull request with the rewriter based method you proposed and I think it definitely seems to be the most elegant solution.

stephencelis:
Thanks!

> While doing the changes I was wondering why the `AllCasePath` declarations default to a public access level instead of reusing the access level of the enum.

We've found this to be the most convenient way to automate the access level, and Apple follows the convention in some of its macros (`@Model`, I believe). We used to match the access level but then we needed to update all our macros when `package` access was introduced, and we've also noticed issues around parent contexts (like `private extension`) that the macro is unaware of.

stephencelis (approved):
Looks good!

PR #188, opened 2024-07-31
Fix some DocC warnings.

mbrandonw:
Just trying to clean up some warnings that have accumulated.

### DIFF
--- a/Sources/CasePaths/Documentation.docc/CaseKeyPath.md
+++ b/Sources/CasePaths/Documentation.docc/CaseKeyPath.md
@@ -14,5 +14,4 @@
 ### Partial case paths
 
 - ``PartialCaseKeyPath``
-- ``Swift/PartialKeyPath/callAsFunction(_:)``
 - ``CasePathable/subscript(case:)-8av06``
--- a/Sources/CasePaths/Internal/Deprecations.swift
+++ b/Sources/CasePaths/Internal/Deprecations.swift
@@ -227,7 +227,7 @@ public func XCTUnwrap<Enum, Case>(
         Actual:
           \(String(describing: `enum`))
       """,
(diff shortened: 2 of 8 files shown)


Sources/CasePaths/Documentation.docc/CaseKeyPath.md
```diff
@@ -14,5 +14,4 @@
 ### Partial case paths
 
 - ``PartialCaseKeyPath``
-- ``Swift/PartialKeyPath/callAsFunction(_:)``
```
mbrandonw, line 17: I don't think this actually exists. Is this old?

Sources/CasePaths/Internal/Deprecations.swift
```diff
@@ -227,7 +227,7 @@ public func XCTUnwrap<Enum, Case>(
         Actual:
           \(String(describing: `enum`))
       """,
-      file: file,
+      file: (file),
```
mbrandonw, line 230: I think we want to silence this warning right?

> ⚠️ Parameter 'file' with default argument '#file' passed to parameter 'file', whose default argument is '#filePath'

stephencelis: I _think_ the way to silence this is to stop using `#file` since its behavior changes in Swift 6 language mode and instead use `#filePath explicitly.

Sources/CasePaths/Documentation.docc/CaseKeyPath.md
stephencelis: Looks like it moved to `_AppendKeyPath` to properly work. Maybe doesn't need to be documented though.

stephencelis (approved):

PR #190, opened 2024-08-07
Case Paths + Key Paths = Optional Paths

stephencelis:
This PR adds some basic support for case/key path composition _via_ "optional" paths.

Often there are case path APIs that don't need the full power of a case path. You might need to extract an optional value, or update an existing optional value, but you don't need to embed that value into a whole new root. This is the behavior of optional-chaining in Swift, which is sadly unsupported when it comes to key paths: an optional-chained key path loses writability.

This PR changes that by expanding the functionality of case key paths so that they gracefully degrade to an "optional" key path.

For example:

```swift
@CasePathable
enum Destination {
  case login(Login)
}

struct Login {
  var username = ""
  var password = ""
}

let loginKeyPath: CaseKeyPath<Destination, Login> = \.login

loginKeyPath as OptionalKeyPath  // ✅

let loginPasswordKeyPath: OptionalKeyPath<Destination, String> = \.login.username  // ✅
```

Some details and experimentation to figure out before merging, so opening as a draft for now.

### DIFF
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -50,14 +50,18 @@ public struct Case<Value> {
   fileprivate let _get: (Any) -> Value?
   fileprivate let _set: Setter
 
+  // NB: Force ABI visibility to avoid aggressive downstream WMO pruning
+  @usableFromInline
   enum Setter {
     case _embed((Value) -> Any)
     case _set((inout Any, Value) -> Void)
 
+    @usableFromInline
     static func embed<Root>(_ embed: @escaping (Value) -> Root) -> Self {
       ._embed(embed)
     }
 
+    @usableFromInline
     static func set<Root>(_ set: @escaping (inout Root, Value) -> Void) -> Self {
       ._set {
         var root = $0 as! Root
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -350,6 +350,7 @@ extension OptionalKeyPath {
   /// - Parameters:
   ///   - lhs: A case key path.
   ///   - rhs: An enum.
+  @_disfavoredOverload
   public static func ~= <Enum: CasePathable, AssociatedValue>(
     lhs: KeyPath,
     rhs: Enum
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -321,44 +321,6 @@ extension CaseKeyPath {
 }
 
 extension OptionalKeyPath {
-  /// Whether an argument matches the case key path's case.
-  ///
-  /// ```swift
-  /// @CasePathable enum UserAction {
-  ///   case settings(SettingsAction)
-  /// }
-  /// @CasePathable enum SettingsAction {
-  ///   case store(StoreAction)
-  /// }
-  /// @CasePathable enum StoreAction {
-  ///   case subscribeButtonTapped
-  /// }
-  ///
-  /// switch userAction {
-  /// case \.settings.store.subscribeButtonTapped:
-  /// // ...
-  /// }
-  ///
-  /// // Equivalent to:
-  ///
-  /// switch userAction {
-  /// case .settings(.store(.subscribeButtonTapped)):
-  /// // ...
-  /// }
-  /// ```
-  ///
-  /// - Parameters:
-  ///   - lhs: A case key path.
-  ///   - rhs: An enum.
-  @_disfavoredOverload
-  public static func ~= <Enum: CasePathable, AssociatedValue>(
-    lhs: KeyPath,
-    rhs: Enum
-  ) -> Bool
-  where Root == Case<Enum>, Value == Case<AssociatedValue> {
-    rhs[case: lhs] != nil
-  }
-
   public func extract<R, V>(from root: R) -> V? where Root == Case<R>, Value == Case<V> {
     Case(self)._extract(from: root)
   }
--- a/Sources/CasePaths/CasePathable.swift
+++ b/Sources/CasePaths/CasePathable.swift
@@ -40,6 +40,16 @@ public protocol CasePathable {
 
   /// A collection of all case paths of this type.
   static var allCasePaths: AllCasePaths { get }
+
+  /// Returns the case name for a given case key path, if available.
+  ///
+  /// - Parameter keyPath: A partial case key path.
+  /// - Returns: The name of the case, or `nil` if the key path doesn't match a known case.
+  static func caseName(for keyPath: PartialCaseKeyPath<Self>) -> String?
+}
+
+extension CasePathable {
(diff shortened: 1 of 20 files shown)


Sources/CasePaths/CasePathable.swift
```diff
@@ -49,60 +49,99 @@ public protocol CasePathable {
...
+        },
+        get: { [$keyPath] in
+          extract(from: $0).flatMap(Value.allCasePaths[keyPath: $keyPath.wrappedValue].extract)
+        },
+        set: { [$keyPath] in
+          set(into: &$0, Value.allCasePaths[keyPath: $keyPath.wrappedValue].embed($1))
+        }
+      )
+    }
+    set {}
```
stephencelis, line 115: The type system now tracks case key paths as _writable_ key paths, and optional paths as plain ole key paths.

```swift
CaseKeyPath    <R, V> = WritableKeyPath<Case<R>, Case<V>>
OptionalKeyPath<R, V> =         KeyPath<Case<R>, Case<V>>
```

This empty setter is never really invoked, it's just responsible for ensuring case key paths compose together without losing embed functionality.

Sources/CasePaths/CasePathable.swift
```diff
@@ -49,60 +49,99 @@ public protocol CasePathable {
...
-  // #endif
 
   public func embed(_ value: Value) -> Any {
     self._embed(value)
   }
 
   public func extract(from root: Any) -> Value? {
-    self._extract(root)
+    self._get(root)
+  }
```
stephencelis, line 141: There's a bit of a mismatch in this PR where sometimes we say "extract" and sometimes we say "get" (mostly just in the initializer for symmetry with "set"). Not sure if we want to consider something else.

Sources/CasePaths/CasePathable.swift
```diff
@@ -277,28 +320,44 @@ extension CaseKeyPath {
   where Root == Case<Enum>, Value == Case<AssociatedValue> {
     rhs[case: lhs] != nil
   }
+
+  public func extract<R, V>(from root: R) -> V? where Root == Case<R>, Value == Case<V> {
+    Case(self).extract(from: root)
+  }
```
stephencelis, line 326: Without a `CasePathable` requirement to tie a `[case:]` subscript to, we need a way to utilize this functionality more generally.

Sources/CasePaths/CasePathable.swift
```diff
@@ -277,28 +320,44 @@ extension CaseKeyPath {
...
+
+  public func set<R, V>(into root: inout R, _ value: V) where Root == Case<R>, Value == Case<V> {
```
stephencelis, line 328: New functionality. Worth bike-shedding the API or good as is?

Sources/CasePaths/CasePathable.swift
```diff
@@ -277,28 +320,44 @@ extension CaseKeyPath {
...
     func open<AnyAssociatedValue>(_ value: AnyAssociatedValue) -> Enum? {
       (Case<Enum>()[keyPath: self] as? Case<AnyAssociatedValue>)?.embed(value) as? Enum
         ?? (Case<Enum>()[keyPath: self] as? Case<AnyAssociatedValue?>)?.embed(value) as? Enum
     }
     return _openExistential(value, do: open)
   }
+
+  public func extract<R>(from root: R) -> Any? where Self == PartialOptionalKeyPath<R> {
+    (Case<R>()[keyPath: self] as? any _AnyCase)?._extract(from: root)
+  }
```
stephencelis, line 360: - [ ] Add a `set(into:_:)` API for partial paths?

Sources/CasePaths/Never+CasePathable.swift
```diff
@@ -16,21 +16,27 @@ extension Case where Value: CasePathable {
   /// This property can chain any case path into a `Never` value, which, as an uninhabited type,
   /// cannot be embedded nor extracted from an enum.
   public var never: Case<Never> {
-    @Sendable func absurd<T>(_: Never) -> T {}
-    return Case<Never>(embed: absurd, extract: { (_: Value) in nil })
+    get {
+      @Sendable func absurd<T>(_: Never) -> T {}
+      return Case<Never>(embed: absurd, extract: { (_: Value) in nil })
+    }
+    set {}
```
stephencelis, line 23: These sets are to preserve case key path functionality. Luckily the weirdness is all hidden away in the "internal" `Case` type.

stephencelis:
A few other things come to mind:

1. Is it weird to call the internal thing `Case` still now that it can refer to non-enum values?
2. While `\Enum.Cases.case` made sense for qualifying case key paths, it makes less sense for optional paths: `\Struct.Cases.property`
3. It's possible to create an "optional" path from any non-optional key path. This is required for composition to work (`\.self` paths, as well as `\.nonOptionalProperty.optionalProperty`). It may feel weird to allow an API that takes optional paths to take a non-optional one, but it also seems the most flexible and we've even had an example in isowords where we wanted to generically `ifLet` to an optional path in one case and a non-optional path in another.

JimRoepcke:
@stephencelis I was using the old OptionalPath code in a project that was on TCA 1.10.4 and swift-case-paths 1.3.0. Trying to migrate it to TCA 1.13.0, and had some compilation errors due to deprecations with case paths.

Just stumbled across this PR, what great timing! Just curious if you know what the timing might be for getting this merged and released?

stephencelis:
Hi @JimRoepcke! Theoretically this branch is ready to go, but we haven't really prioritized it. In most of our projects we've remodeled our domains in ways that no longer require optional paths, and so the main thing needed for this branch to land is some beta testers, which is where you could come in :smile:

Want to point your project at this branch and take things for a spin to see if it unblocks you? We're eager for feedback from users with real world requirements!

JimRoepcke:
@stephencelis I pointed my code at this branch. Working through issues in my code, but I'm also seeing TCA itself is not building?

image

fruitcoder:
Hi @stephencelis, will this also fix an issue I had when trying to model my domain by wrapping my enum reducer in another enum before putting it in State? The code can be simplified to this

```swift
@CasePathable
enum LoadingState: Equatable {
  case loading
  case loaded(ContentReducer.State)
}

@Reducer(state: .equatable)
enum ContentReducer {
  case child1(Child1Reducer.State)
}

@Reducer
struct Child1Reducer {
  typealias State = Never
  typealias Action = Never
  typealias Body = EmptyReducer
}

@Reducer
struct Child2Reducer {
  typealias State = Never
  typealias Action = Never
  typealias Body = EmptyReducer
}

@Reducer
struct Feature {
  @ObservableState
  struct State: Equatable {
    var content: LoadingState = .loading
  }

  enum Action {
    case content(ContentReducer.Action)
  }

  public var body: some ReducerOf<Self> {
    EmptyReducer()
      .ifLet(\.content.loaded, action: \.content) {
        ContentReducer.body
      }
  }
}
```
This gives me the following error:
image

stephencelis:
@fruitcoder Technically it should, though TCA would need to update all its helpers to work with optional paths rather than key paths and case paths.

PR #191, opened 2024-08-15
Fix deprecation warning

finestructure:


stephencelis (approved):
Think we can merge this now that folks should no longer be on the beta :smile: Thanks, Sven!

PR #193, opened 2024-08-28
Follow-up Wasm CI integration

kateinoigakukun:
We are moving to Swift SDK-style installation on the development branch, and `swift test` with static stdlib is broken in obsoleting toolchain-style installation now.

Verified it works here ✅  https://github.com/kateinoigakukun/swift-case-paths/actions/runs/10591664518/job/29349506477


mbrandonw (approved):
This is awesome, thanks so much @kateinoigakukun! We'll try to get Wasm CI on all of our repos that it's possible.

PR #196, opened 2024-09-06
fix naming consistancy accross pointfree libs

doozMen:
There is an issue when using case-paths and
When you already use the swift-issue-reporting dependency by using the url for swift-issue reporting.

> I wrongly described the issue to be snapshot testing. I was having an unrelated issue with that and confused it here. 

The issue is that in swift 6 a swift package containing a reference to the same package but with different url results in an error where this previously was a warning. 
So the name can stay but could you change the url? 

Or you could also close my PR as simply removing my dependency on issue reporting and reusing the one from case-path worked too

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,4 +1,5 @@
 {
+  "originHash" : "1d5b1b96f9e9f04f52a6effd94db47d675cb0faea4b3846b36b610fd4a8c05ed",
   "pins" : [
     {
       "identity" : "swift-docc-plugin",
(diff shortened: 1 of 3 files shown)


stephencelis:
@doozMen Snapshot Testing doesn't depend on Issue Reporting at all:

https://github.com/pointfreeco/swift-snapshot-testing/blob/main/Package.swift

Can you explain the conflict you're seeing and share a project that demonstrates the issue?

We currently lock to `xctest-dynamic-overlay` as the package name because it's the only way to support things in our package ecosystem in a backwards-compatible way. In the future we will split these two packages out to better address the issue, but unfortunately do not think we can merge this PR without breaking things for folks.

doozMen:
I updated the description. I mixed 2 unrelated issues I had I'm afraid. Please let me know if this is clear enough or if I should make a small package that reproduces the confusing issue whether using swift 6

stephencelis:
@doozMen For now you must specify `xctest-dynamic-overlay` to work around the problem. In the future we will be able to fix this by splitting the package into 2.

PR #197, opened 2024-09-20
[Proposal] Add Swift Macro Compatibility Check GitHub Action

Matejkob:
This PR adds a **Swift Macro Compatibility Check** GitHub Action to ensure that macro changes remain compatible across different versions of `swift-syntax`. The action automatically verifies compatibility with major `swift-syntax` versions (`509.0.0`, `510.0.0`, `600.0.0`), running builds and tests. It's also posible to turn on checking agains all versions of `swift-syntax`.

#### Key Benefits:
- **Automated Compatibility Checks**: No need to manually test changes across different `swift-syntax` versions—this is handled automatically in CI.
- **Prevents Breakages**: Helps avoid accidentally committing changes that break older versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,18 @@ jobs:
       - name: Run tests
         run: wasmtime --dir . .build/debug/swift-case-pathsPackageTests.wasm
 
+  check-macro-compatibility:
+    name: Check Macro Compatibility
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Run Swift Macro Compatibility Check
+        uses: Matejkob/swift-macro-compatibility-check@v1.0.0
+        with:
+          run-tests: false
+          major-versions-only: false
+
   # windows:
   #   name: Windows (Swift ${{ matrix.swift }}, ${{ matrix.config }})
   #   strategy:


Matejkob:
@mbrandonw @stephencelis I’d love to hear your thoughts on this! I've been using such a script in my internal macros as well as my open-source macro [swift-spyable](https://github.com/Matejkob/swift-spyable) for a while and it's huge improvment. Today, finally, I have managed to find time to make a GitHub Action out of it that can be used across diffrent repos to ensure that all macros are good swift-syntax citizens ;) I think integrating this action would be a nice first step forward in ensuring compatibility with multiple versions of swift-syntax in the whole ecosystem in general.

In my opinion, this addition adds a layer of safety without introducing much overhead since the process is fully automated. Plus, it provides quick feedback on any potential issues.

What do you think? Do you see this as a positive step? I'd love to discuss any concerns or suggestions you may have!

stephencelis:
@Matejkob Cool! @mbrandonw and I will discuss early next week :smile:

stephencelis (approved):
We're gonna give this a shot! A main worry we had was CI time, but it doesn't take an exorbitant amount of time, and hopefully it'll save us sometime in the future.

If it every breaks in the future we may disable and ping you, but for now let's do it!

Matejkob:
> We're gonna give this a shot! A main worry we had was CI time, but it doesn't take an exorbitant amount of time, and hopefully it'll save us sometime in the future.

If we don't run tests—only build the project—the time spent is very reasonable, as running the `clean` action isn't required between each iteration. The current setup checks against all versions of `swift-syntax`, but I've noticed that the impact compared to only checking major versions is not significant in terms of CI usage. 

Speaking of saving time in the future, it has already saved me twice, so hopefully, it will be useful to you as well.

> If it every breaks in the future we may disable and ping you, but for now let's do it! 

Sure thing, thanks!

PR #198, opened 2024-09-23
Update dependencies, to avoid build issues with Xcode 16 on `make test-swift`

gemmakbarlow:
Previously, running `make test-swift` would result in build errors with Xcode 16 selected, i.e. 

image

Updated several dependencies to resolve this, and have the command complete successfully. 

Additionally, updated CI to ensure Xcode 16 is being used across actions.


stephencelis (approved):
Thanks for the CI updates!

PR #202, opened 2024-09-29
Addresses build issues seen when building for library evolution, Xcode 16

gemmakbarlow:
Addresses some build issues with Xcode 16 when building for Library evolution, i.e.

image

These can be seen by reverting the change in Sources/CasePathsMacros/CasePathableMacro.swift and running `make build-for-library-evolution`

Adds CI around the new `make` target, for future-proofing.


gemmakbarlow:
Whoops, forgot I already PRed this 🤦🏻‍♀️

PR #204, opened 2024-10-01
Added a convenience `is` function for optional CasePathable values

HarshilShah:
This PR is aimed at improving ergonomics for `is` checks specifically in TCA. Destinations and other pieces of CasePathable state can often be optional in TCA, and so out of the box the `CasePathable.is` function requires either unwrapping the optional state or tacking on `== true` to compile. I defined this extension in my own codebase a while ago to make that syntax a bit nicer to read, so as I was copying it over to another project I thought it might be worth submitting for inclusion right in the library itself

### DIFF
--- a/Sources/CasePaths/Optional+CasePathable.swift
+++ b/Sources/CasePaths/Optional+CasePathable.swift
@@ -88,23 +88,8 @@ extension Optional.AllCasePaths: Sequence {
 }
 
 extension Optional where Wrapped: CasePathable {
-  /// Tests the associated value of an optional case.
-  ///
-  /// ```swift
-  /// @CasePathable
-  /// enum UserAction {
-  ///   case home(HomeAction)
-  ///   case settings(SettingsAction)
-  /// }
-  ///
-  /// let userAction: UserAction? = .home(.onAppear)
-  /// userAction.is(\.home)      // true
-  /// userAction.is(\.settings)  // false
-  ///
-  /// let userActions: [UserAction] = [.home(.onAppear), .settings(.subscribeButtonTapped)]
-  /// userActions.filter { $0.is(\.home) }      // [UserAction.home(.onAppear)]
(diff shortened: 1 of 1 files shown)


Sources/CasePaths/Optional+CasePathable.swift
```diff
@@ -86,3 +86,26 @@ extension Optional.AllCasePaths: Sequence {
...
+  /// }
+  ///
+  /// let userAction: UserAction? = .home(.onAppear)
+  /// userAction.is(\.home)      // true
+  /// userAction.is(\.settings)  // false
+  ///
+  /// let userActions: [UserAction] = [.home(.onAppear), .settings(.subscribeButtonTapped)]
+  /// userActions.filter { $0.is(\.home) }      // [UserAction.home(.onAppear)]
+  /// userActions.filter { $0.is(\.settings) }  // [UserAction.settings(.subscribeButtonTapped)]
+  /// ```
```
stephencelis, line 107: Since `Optional` _is_ `CasePathable` I think we can hide this overload's primary use as an aid in autocomplete.

```suggestion
  @_disfavoredOverload
  @_documentation(visibility:internal)
```

stephencelis (approved):
Hey @HarshilShah! Thanks for opening the PR as it made a deficiency clear to me :smile:

_Technically_ this API isn't actually needed, because `Optional` _is_ `CasePathable`, and because you can squash `\.some` chaining when forming case key paths. You can confirm this by compiling the following today:

```swift
@CasePathable enum Foo { case bar }

func f() {
  let optional = Optional(Foo.bar)
  optional.is(\.bar)
}
```

I think the problem is that Xcode autocomplete will favor optional-chaining into the `Wrapped` over applying the method directly, so you get the following by default:

```swift
optional?.is(\.bar)
```

So I think this overload is worth adding to the library if only to make the autocomplete experience better :smile:

Thanks again!

PR #211, opened 2024-12-21
Use custom GitHub Action for Android

johankool:
Ran successfully [here](https://github.com/johankool/swift-case-paths/actions/runs/12446523808/job/34748996196)


stephencelis (approved):
Great!

PR #212, opened 2025-01-09
Add `CasePathsCore` module

stephencelis:
The `@CasePathable` macro requires SwiftSyntax, which is quite heavy to rely on and build, but not every library or app depending on case paths needs to build or load the macro, so let's extract the core functionality of case paths to a "core" module that can be built and linked against in isolation.


mbrandonw (approved):

stephencelis:
To build combined documentation:

```
swift package generate-documentation --enable-experimental-combined-documentation --target CasePaths --target CasePathsCore
```

mbrandonw (approved):

PR #221, opened 2025-04-09
Use IssueReporting and XCTestDynamicOverlay from swift-issue-reporting

philipbel:
swift-case-paths uses xctest-dynamic-overlay.  This causes issues (with Tuist) when mixing packages that use SwiftIssueReporting with swift-case-paths.

As far as I understand, SwiftIssueReporting supersedes xctest-dynamic-overlay and hence the redirect.

I ran `swift test`.  I see some of the deprecated tests fail, both before and after my change.  I assume that's OK?  Can include output if needed.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "75a28c180eaa852896a1e043cbbf340d2f28af3ec620072b1a1b7d9a6041d14e",
+  "originHash" : "4ac7d158042932865cdeb2bf41e244cdc8e479f97add87a79a2244304dcdff2a",
   "pins" : [
     {
       "identity" : "swift-docc-plugin",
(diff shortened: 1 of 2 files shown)


mbrandonw:
Hi @philipbel, we have [this note](https://github.com/pointfreeco/swift-issue-reporting?tab=readme-ov-file#overview) in the readme of Issue Reporting:

> Important
>
> Issue Reporting is an evolution of our previous library, XCTestDynamicOverlay. As such, to use this library you must depend on the old repository URL:
> 
> https://github.com/pointfreeco/xctest-dynamic-overlay

For the time being it is not correct to depend on pointfreeco/swift-issue-reporting. You must depend on pointfreeco/xctest-dynamic-overlay and use the IssueReporting module from that library.

If you have any projects depending on pointfreeco/swift-issue-reporting simply update them to point to pointfreeco/xctest-dynamic-overlay instead and everything should work just fine.

PR #223, opened 2025-05-15
Add CasePathsCore usage example

klaaspieter:
I was looking for an explanation on how to use  `CasePathsCore` and happened to found it in the `CasePathable` documentation comment. 

Before I found it though I wasn't aware that `CasePathable` was the protocol the macro synthesized conformance for. This moves the information to a place that is easier to find. Although maybe it should even be in a CasePathsCore README?

### DIFF
--- a/Sources/CasePathsCore/Documentation.docc/CasePathsCore.md
+++ b/Sources/CasePathsCore/Documentation.docc/CasePathsCore.md
@@ -10,6 +10,40 @@ macro, and is automatically imported when you `import CasePaths`
 See the [`CasePaths`](../casepaths) module for information about the `@CasePathable` macro and
 other non-core functionality.
 
+To use Case paths without relying on the `@CasePathable` macro import `CasePathsCore` and manually conform your type to the `CasePathable` protocol.
+
+For example the `Result` type is extended to be case-pathable with the following extension:
+
+```swift
+import CasePathsCore
+
+extension Result: CasePathable {
+  public struct AllCasePaths {
+    var success: AnyCasePath<Result, Success> {
+      AnyCasePath(
+        embed: { .success($0) },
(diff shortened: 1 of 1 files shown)


stephencelis:
I might not be understanding the motivation behind this PR. The docs you added appear to come straight from the `CasePathable` protocol documentation. 

Can you clarify the problem you were having? I'm not sure I understand the motivation behind "how to use `CasePathsCore`". Did you encounter the module name when adding CasePaths to an Xcode project and wanted to understand the difference between `CasePathsCore` and `CasePaths`? Or did you find `CasePathsCore` a different way?

If it's a question of documenting the modules for installation, we recently added this to our SharingGRDB project: https://github.com/pointfreeco/sharing-grdb/pull/49/files

Would such a tip have helped you here?

klaaspieter:
Yeah, such a tip would help. I was basically looking for how to use `CasePathsCore` without the `CasePathable` macro and I had to go hunting for what the macro actually did to know that I was looking for the `CasePathable` protocol and it's documentation. In other words I think if you're coming to this library without any knowledge about how it works it's difficult to find the necessary information to use it without the macro.

You're right that my PR merely lifts the existing documentation to a place that I think is easier to find. I was expecting this (or similar) information to be in the `CasePathsCore` README but decided to update the documentation instead because that already exists 🤷.

stephencelis:
@klaaspieter Sorry for dragging on this, but the good news is that with SwiftSyntax prebuilts the `CasePaths`-`CasePathsCore` dichotomy is no longer necessary, and in #235 will be removed entirely. Because of this I think we can close this PR out. Let me know if you have any other questions or concerns!

klaaspieter:
No worries. Thanks for the heads up!

PR #224, opened 2025-06-14
Add support for Swift 6.1 trailing commas

brzzdev:
Turns out the [issue](https://github.com/pointfreeco/swift-composable-architecture/issues/3709) was actually coming from swift-case-paths.

I would've added tests, but the MacroTests don't seem to be building. Not sure if that's a me problem, or a know problem.


brzzdev:
@mbrandonw Added!

Like I mentioned, I wanted to add them initially but I was getting errors building the tests last night. A `Reset Package Caches` seemed to fix it.

mbrandonw (approved):

PR #237, opened 2025-11-24
Include Missing Target Dependencies

CrownedPhoenix:
Adds missing target dependencies for a few imports within CasePathsMacros.
I was getting build failures due to these issues on Swift 6.2 SPM (`swift build`).

### DIFF
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -47,6 +47,8 @@ let package = Package(
         .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
         .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
         .product(name: "SwiftDiagnostics", package: "swift-syntax"),
+        .product(name: "SwiftSyntax", package: "swift-syntax"),
+        .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
       ]
     ),
     .testTarget(
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -46,6 +46,7 @@ let package = Package(
       dependencies: [
         .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
         .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
+        .product(name: "SwiftDiagnostics", package: "swift-syntax"),
       ]
     ),
     .testTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -47,6 +47,8 @@ let package = Package(
         .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
         .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
         .product(name: "SwiftDiagnostics", package: "swift-syntax"),
+        .product(name: "SwiftSyntax", package: "swift-syntax"),
+        .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
       ]
     ),
     .testTarget(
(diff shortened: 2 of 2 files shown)


Package@swift-5.9.swift
```diff
@@ -71,6 +74,7 @@ if ProcessInfo.processInfo.environment["OMIT_MACRO_TESTS"] == nil {
       name: "CasePathsMacrosTests",
       dependencies: [
         "CasePathsMacros",
+        .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
```
CrownedPhoenix: This addition is a bit weird to me. I'm not sure why swift complained about this since I would have expected it to get this module transitively through `CasePathsMacros` but I'm wondering if the SwiftCompilerPlugin has some unusual behavior given the place that it sits in the compile  pipeline. 🤷

CrownedPhoenix: ~~This seems related to: https://github.com/swiftlang/swift-syntax/issues/3188
Happy to drop this commit if desired.~~

I was mistaken. This was something else.

mbrandonw: Hi @CrownedPhoenix, thanks for looking into this. The other changes make sense to me, but this one does not. Can you double check that the other changes alone do not work to get your project into compiling order? If they do work, then I think we can undo this. And if it doesn't work, would you be able to share a project that reproduces the problem for us to look into?

CrownedPhoenix: I (locally) reverted this commit and nuked the `.build` folder.
I'm using:
```
> swift --version
Apple Swift version 6.2 (swift-6.2-RELEASE)
Target: arm64-apple-macosx15.0
Build config: +assertions
```
I'm running Sequioia 15.7.1 on an M4 Pro.

and I ran: `swift build --build-tests`.

I get the following error:
```
error: emit-module command failed with exit code 1 (use -v to see invocation)
~/repos/CrownedPhoenix/swift-case-paths/Tests/CasePathsMacrosTests/CasePathableMacroTests.swift:2:10: error: missing required module 'SwiftCompilerPlugin'
   1 | #if canImport(MacroTesting)
   2 |   import CasePathsMacros
     |          `- error: missing required module 'SwiftCompilerPlugin'
   3 |   import MacroTesting
   4 |   import SwiftSyntaxMacros
~/repos/CrownedPhoenix/swift-case-paths/Tests/CasePathsMacrosTests/CasePathableMacroTests.swift:2:10: error: missing required module 'SwiftCompilerPlugin'
   1 | #if canImport(MacroTesting)
   2 |   import CasePathsMacros
     |          `- error: missing required module 'SwiftCompilerPlugin'
   3 |   import MacroTesting
   4 |   import SwiftSyntaxMacros
```

I was able to reproduce on the swift:6.2 official docker image.
Check out: https://github.com/CrownedPhoenix/swift-case-paths/tree/experiment/swiftcompilerplugin-failure and run `docker compose up` to reproduce (🤞 )

~~EDIT: I just saw `reproducer-1  | error: Dependencies could not be resolved because no versions of 'swift-macro-testing' match the requirement 0.2.0..<1.0.0 and root depends on 'swift-macro-testing' 0.2.0..<1.0.0.` when I re-tried my docker compose up so I might have something wrong with the setup on that branch. Looking at it now.~~

Nope it's fine something was up with my cache. The reproducer should work.

mbrandonw: Hi @CrownedPhoenix, how about we revert this line for this PR so that we can merge and release it right away. And then it seems like an issue should be opened on SwiftSyntax to see why is this extra dependency required. We are worried that if we take it it will fix your issue and then possibly cause problems for someone else.

CrownedPhoenix: Sounds good. I'll drop this commit. Thanks for the feedback! 😄

stephencelis (approved):

stephencelis:
@CrownedPhoenix Thanks! As for the other explicit test dependency on `SwiftCompilerPlugin`, please let us know of any issue/discussion you start on SwiftSyntax so we can follow it.

PR #238, opened 2025-12-07
Add swift-parser-diagnostics as direct dependency

paulgessinger:
Starting in XCode 26.2 I get warnings like:

```
'CasePathsMacros' is missing a dependency on 'SwiftParserDiagnostics' because dependency scan of Swift module 'CasePathsMacros' discovered a dependency on 'SwiftParserDiagnostics'
```

I think this is because indeed that target is not explicitly listed. This PR is my attempt to fix that.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -49,6 +49,7 @@ let package = Package(
         .product(name: "SwiftDiagnostics", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
+        .product(name: "SwiftParserDiagnostics", package: "swift-syntax"),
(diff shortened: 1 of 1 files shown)


SplittyDev:
I think it's a transient issue. Cleaning the build cache and rebuilding made it work for me on Xcode 26.2

paulgessinger:
@SplittyDev I was on the 26.2 RC and a clean build didn't seem to make a difference for me. But I'll try again on the released version, and report back.

stephencelis:
@paulgessinger Circling back around to this. Is this still necessary in your project, and if so can you share a project that reproduces?

stephencelis:
@paulgessinger Gonna close this for now, but if it is still an issue we'd love a repro in a new issue!

PR #239, opened 2026-03-30
Support swift-syntax 605

shimastripe:
Bumps swift-syntax upper bound from 603 to 605 to support 603.x and 604.x prereleases.
https://github.com/search?q=org%3Apointfreeco+https%3A%2F%2Fgithub.com%2Fswiftlang%2Fswift-syntax+603.0.0&type=code

- same: https://github.com/pointfreeco/swift-dependencies/pull/416

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ let package = Package(
     ),
   ],
   dependencies: [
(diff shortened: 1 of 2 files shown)


stephencelis (approved):
Thanks!

PR #255, opened 2026-08-27
Add `CasePathsCore.CasePath` protocol

stephencelis:
This will aid in libraries that want to integrate with both CasePaths 1.0 and 2.0. They will be able to make the following change today:

```diff
 import CasePathsCore

-KeyPath<Root.AllCasePaths, AnyCasePath<Root, Value>>
+KeyPath<Root.AllCasePaths, some CasePath<Root, Value>>
```

And 2.0 should work without any changes.

### DIFF
--- a/Sources/CasePaths/Internal/Deprecations.swift
+++ b/Sources/CasePaths/Internal/Deprecations.swift
@@ -294,14 +294,6 @@ public func XCTModify<Enum, Case>(
 
 // Deprecated after 1.0.0:
 
-/// A type-erased case path that supports embedding a value in a root and attempting to extract a
-/// root's embedded value.
-///
-/// This type has been renamed to `AnyCasePath` and is primarily employed by the ``CasePathable()``
-/// macro to derive `CaseKeyPath`s from an enum's cases.
-@available(*, deprecated, renamed: "AnyCasePath")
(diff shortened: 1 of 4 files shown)


Sources/CasePaths/Internal/Deprecations.swift
```diff
@@ -291,14 +291,6 @@ public func XCTModify<Enum, Case>(
 
 // Deprecated after 1.0.0:
 
-/// A type-erased case path that supports embedding a value in a root and attempting to extract a
-/// root's embedded value.
-///
-/// This type has been renamed to `AnyCasePath` and is primarily employed by the ``CasePathable()``
-/// macro to derive `CaseKeyPath`s from an enum's cases.
-@available(*, deprecated, renamed: "AnyCasePath")
-public typealias CasePath = AnyCasePath
```
stephencelis, line 300: While we could keep this alive and have dependents do a scoped import, it does require using old compiler features that aren't well-supported today. This deprecated alias was introduced in 1.0.0 for migration's sake only, back in August 2023, and was never meant to be used otherwise, so it should be safe to prune as we prepare for 2.0.

mbrandonw (approved):

